PR #3, opened 2022-02-27
logs

o-volkov:


bombers.tf
```diff
@@ -0,0 +1,131 @@
...
+    source      = "main"
+    destination = "/tmp/main"
+  }
+  provisioner "remote-exec" {
+    inline = [
+      "chmod +x /tmp/main",
+      "/tmp/main"
+    ]
+  }
+}
```
Arriven, line 50: Can we  pleasemove this to https://github.com/db1000n-coordinators/db1000n-infra ?

main.go
```diff
@@ -120,16 +120,22 @@ func httpJob(ctx context.Context, args JobArgs) error {
 		for key, value := range jobConfig.Headers {
 			req.Header.Add(parseStringTemplate(key), parseStringTemplate(value))
 		}
+
+		startedAt := time.Now().Unix()
+		log.Printf("[DEBUG] %s %s started at %d", jobConfig.Method, jobConfig.Path, startedAt)
```
Arriven, line 125: It would be ideal to use some more sophisticated logging library that supports logging levels so that we could only include debug logging when needed. Can be done in a separate PR if you want

Arriven (commented):
Overall looks like a good change but left two minor comments

PR #24, opened 2022-02-27
new relic

zmitry:
- add docker readme
- add new relic metrics


Arriven:
Decided to go forward with google analytics as it's cheaper and is more in line with the current needs

PR #74, opened 2022-03-04
Create VPN tunnel from within Docker container

bitshape:
Pros of this approach:

* You can place as many `*.ovpn` configurations into `openvpn` directory to rotate between them randomly
* If container's VPN dies for any reason, it will take the whole container with it meaning less chance of IP leakage (further hardening could be achieved with `iptables` if needed)

Downsides:

* Since VPN liveness is verified by `supervisord`, we need to change entry point for the container which means it's not possible to run this interactively like `db1000n -c http://someother.url`


Dockerfile
```diff
@@ -9,7 +9,18 @@ RUN CGO_ENABLED=0 GOOS=linux go build -a -o main .
 
 FROM alpine:3.11.3
 
+RUN apk add --update supervisor openvpn curl && rm  -rf /tmp/* /var/cache/apk/*
```
Arriven, line 12: This is a very cool idea and setup!
Can you move it to a separate Dockerfile or separate build target in current Dockerfile so that I could make a separate package for it and update instructions? Ideally I want people to have as much options as possible

Arriven: I'll merge and try to achieve that closer to the evening myself if you don't get a chance to do this simply since the idea is very good

Arriven:
@bitshape pls tag me here or via other channels if you get a chance to address my comment

Arriven:
Cherry picked it into main branch and applied the fix to change this config to be non-default target

PR #93, opened 2022-03-04
Minor cleanup; move templates to a separate file; prepare for splitti…

jdoe7865623:
…ng template parsing and execution


template.go
```diff
@@ -0,0 +1,90 @@
+package main
```
Arriven, line 1: Can you move this into utils subpackage? I want to start moving some code out of the main file to improve the overall structure so it's probably better to try applying that rule for all new code

main.go
```diff
@@ -196,14 +121,13 @@ func httpJob(ctx context.Context, l *logs.Logger, args JobArgs) error {
...
 	var jobConfig httpJobConfig
-	err := json.Unmarshal(args, &jobConfig)
-	if err != nil {
+	if err := json.Unmarshal(args, &jobConfig); err != nil {
 		l.Error("error parsing json: %v", err)
 		return err
 	}
+
 	var clientConfig HTTPClientConfig
-	err = json.Unmarshal(parseByteTemplate(jobConfig.Client), &clientConfig)
```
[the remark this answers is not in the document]
jdoe7865623: You're right, but I plan to get rid of it altogether so that there is one function for template parsing, other for execution. This is temporary.

template.go
jdoe7865623: I thought about the same thing. We can move every job into a separate file first, split job functions into smaller ones for readability and then think about subpackaging.

PR #114, opened 2022-03-04
Added country check to make sure users don't launch it with Ukrainian IPs

### LINKED ISSUE #110: Check if user enabled VPN and so whether attack will be effective
If user's IP address detected as Ukrainian, open page in browser with instruction on how to setup VPN.
Will try to do myself.

Devaniti:
First time coding in go, so code may be horrible
Need to replace URL to correct one
closes #110

### DIFF
--- a/main.go
+++ b/main.go
@@ -141,9 +141,9 @@ type IPInfo struct {
 func openBrowser(url string, l *logs.Logger) {
 	switch runtime.GOOS {
 	case "windows":
-		exec.Command("rundll32", "url.dll,FileProtocolHandler", url).Start()
+		_ = exec.Command("rundll32", "url.dll,FileProtocolHandler", url).Start()
 	case "darwin":
-		exec.Command("open", url).Start()
+		_ = exec.Command("open", url).Start()
 	}
 	l.Info("Please open %s", url)
 }
@@ -174,6 +174,7 @@ func main() {
 	if err != nil {
 		l.Warning("Can't check users country. Please manually check that VPN is enabled or that you have non Ukrainian IP adress.")
 	} else {
+                defer resp.Body.Close()
 		body, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
 			l.Warning("Can't check users country. Please manually check that VPN is enabled or that you have non Ukrainian IP adress.")
--- a/main.go
+++ b/main.go
@@ -29,9 +29,12 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -152,6 +169,31 @@ func main() {
 
 	l := logs.New(logLevel)
 	clientID := uuid.New().String()
+	
+	resp, err := http.Get("https://api.myip.com/")
```
Arriven, line 173: There should be 
```suggestion
	defer resp.Body.Close()
```
After error checking (if I recall correctly you're a c++ dev and it's the way RAII is done for go)

main.go
```diff
@@ -131,6 +134,20 @@ func updateConfig(configPath, backupConfig string) (config *Config, err error) {
...
 }
 
+type IPInfo struct {
+	Country string `json:"country"`
+}
+
+func openBrowser(url string, l *logs.Logger) {
+	switch runtime.GOOS {
+	case "windows":
+		exec.Command("rundll32", "url.dll,FileProtocolHandler", url).Start()
```
Arriven, line 144: (Optional) Please explicitly check for error or explicitly ignore it with
```suggestion
		_ = exec.Command("rundll32", "url.dll,FileProtocolHandler", url).Start()
```

Arriven (commented):
Overall it looks good, only one minor thing to fix in order to avoid resources leaking

main.go
Arriven: Ah, interesting, github even allows you to commit suggestions directly via the web UI. I'll try to check it out right now :)

main.go
```diff
@@ -131,6 +134,20 @@ func updateConfig(configPath, backupConfig string) (config *Config, err error) {
...
+	case "darwin":
+		exec.Command("open", url).Start()
```
Arriven, line 146: testing github suggetsion functionality
```suggestion
		_ = exec.Command("open", url).Start()
```

main.go
Arriven: I need to resolve this first to make a proper suggestion here

main.go
```diff
@@ -152,6 +169,31 @@ func main() {
 
 	l := logs.New(logLevel)
 	clientID := uuid.New().String()
+	
+	resp, err := http.Get("https://api.myip.com/")
+	if err != nil {
+		l.Warning("Can't check users country. Please manually check that VPN is enabled or that you have non Ukrainian IP adress.")
+	} else {
+		body, err := ioutil.ReadAll(resp.Body)
```
Arriven, line 177:
```suggestion
                defer resp.Body.Close()
		body, err := ioutil.ReadAll(resp.Body)
```

Arriven (commented):
Made suggestions to try to commit them via github

PR #130, opened 2022-03-05
Added script for one click start

Devaniti:


Arriven (approved):
Looks good to me, but can you move it into a scripts/ folder please? I'll merge in couple minutes anyway but having better file structure would be nice considering the amount of controbutors

scripts/StartDB1000N.bat
```diff
@@ -0,0 +1,28 @@
+echo $SaveFolder = $args[0]                                                       >  %temp%\GetDB1000N.ps1
```
Amet13, line 1: @Devaniti is there any option to add a short comment in this script, what does it for? For people who are not familiar with Windows?

Devaniti: makes sense

Arriven: Let's do it as a separate PR

PR #135, opened 2022-03-05
Fix some misspells and misslinks in docs

Amet13:
Just went through all docs and fixed misspells and misslinks


5har0varik (approved):
Looks good

PR #138, opened 2022-03-05
Add supervisor for app restart after crash

### LINKED ISSUE #123: [improvement] [win] BAT file with supervisor/guardian
To prevent windows crash add simple supervisor in .bat file. Add .bat file to .zip package. 
For example:
```
echo starting
:start
start /w db1000n.exe
echo restarting
goto start
```

5har0varik:
closes #123

### DIFF
--- a/scripts/StartDB1000N.bat
+++ b/scripts/StartDB1000N.bat
@@ -37,4 +37,8 @@ rem Download latest windows x64 build
 rem We don't currently check if it's already downloaded, so it redownload latest version each run
 powershell -ExecutionPolicy Bypass -File %temp%\GetDB1000N.ps1 %temp%
 rem Assume that archive contained executable file named db1000n.exe and try to run it
-%temp%/db1000n.exe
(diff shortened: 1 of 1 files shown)


Arriven:
@Devaniti can you take a quick look? it looks ok to me but I'd rather check with you until I get enough time to get fluent in that script

Devaniti:
On windows ctrl-c seems to generate non 0 exit code
You can take a look at go code to make it return 0 in that case, and I'll write supervisor in that script in the meantime

Devaniti:
Oh wait, Ctrl+C stops whole batch script, so supervisor will not deadlock even if exit code is wrong

Devaniti:
I made another pull request, same idea, but with comments, without using start command (I guess it's my personal preference), and with graceful exit from script when app does graceful exit
https://github.com/Arriven/db1000n/pull/141
Thanks @5har0varik for the idea!

PR #140, opened 2022-03-05
Implement job runner

jdoe7865623:


src/runner/runner.go
```diff
@@ -0,0 +1,231 @@
...
+
+		l.Info("Loading config from %q", paths[i])
+
+		return res, nil
+	}
+
+	return nil, errors.New("config fetch failed")
+}
+
+func fetchConfigOnePath(path string) ([]byte, error) {
```
Arriven, line 160: Not required for merge, but if you get back to this PR before I get to my laptop - would be nice to rename it to something like `fetchSingleConfig`

src/runner/runner.go
```diff
@@ -0,0 +1,231 @@
...
+	configPaths    []string
+	backupConfig   []byte
+	refreshTimeout time.Duration
+	metricsPath    string
+
+	currentRawConfig []byte // currently applied config
+
+	log *logs.Logger
+
+	stop chan interface{}
```
Arriven, line 44: Also not critical, but it's better to pass `context.Context` instead of reinventing it

Arriven (approved):
LGTM, will merge when I get back to laptop

src/runner/runner.go
jdoe7865623: No probs

src/runner/runner.go
jdoe7865623: I don't agree here. `context.Context` is more expensive and all we need here is `chan` to close and signal shutdown. This is a canonical concurrent operation in go.

PR #151, opened 2022-03-05
Add DNS flood job

### LINKED ISSUE #8: DNS stress
- could integrate https://github.com/MickaelBergem/dnsstresss/blob/master/dnsstresss.go

AndriiChuzhynov:
This is first iteration of DNS flood job.
Based on dnsstresss.go but adapted to work with db1000n.
Each job can be pointed on the resolver with multiple target domains, check testconfig.json
May close #8


dnsflood/dnsflood.go
```diff
@@ -0,0 +1,106 @@
...
+	setDefaults(config)
+	s := &DnsStress{
+		Logger: logger,
+		Config: config,
+	}
+
+	parsedResolver, err := ParseIPPort(config.Resolver)
+	config.Resolver = parsedResolver
+	if err != nil {
+		s.Logger.Error("Unable to parse the resolver address", err)
```
sivillakonski, line 46: As a hint, you can add the address of the resolver for faster debug

AndriiChuzhynov:
Random subdomain may be a good idea https://github.com/galprz/dns-random-subdomains-ddos-attack

Arriven:
@AndriiChuzhynov I went ahead with #166 instead, although I have a feeling that we can have both jobs even if they're doing similar things. We can always remove one of them later once we test both enough but it's up to you if you want to continue with this PR

AndriiChuzhynov:
Hi @Arriven I'm ok with #166 but, I think, we need to add dynamic nameserver resolving. 
Because NS servers IPs could change pretty fast this days

Arriven:
Can you make an issue or PR for that?

AndriiChuzhynov:
closing in favor of #166

PR #154, opened 2022-03-05
Move to default go log package

jdoe7865623:


Arriven:
There are merge conflicts in this. If I recall correctly your main concern is performance, can we maybe resort to switching the package inside logs implementation in order to minimize changes? I'd also like to see at least a minimal comparison on how this impacts performance

PR #160, opened 2022-03-05
add pull policy to docker run in order to use the very last container

iamtodor:


Arriven:
Hmm, I thought I already added that, thanks for catching!

Issue #133, opened 2022-03-05
Network activity went down today.


vgoncharenko:
Guys, something happened around 2:20 UTC.
my command: docker run --pull=always ghcr.io/arriven/db1000n:latest
Unfortunately, cant track if it was config changes.
<img width="1095" alt="Screen Shot 2022-03-05 at 5 36 33 AM" src="https://user-images.githubusercontent.com/10031428/156881337-0681a472-38d2-4447-900a-df349d8e64bf.png">

Arriven:
I don't recall any config changes at that time. Need to investigate

vgoncharenko:
I don't have vpn, btw. 
And there is no packet loss:

eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 9001
        ................................................................................................
        RX packets 1403319  bytes 88604937 (84.5 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 283605121  bytes 1019714276523 (949.6 GiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

uptime:
 12:28:11 up 15:12,  2 users,  load average: 3.19, 2.81, 2.60

Let me know if I can help with any info.

Arriven:
I don't have any ideas on possible cause yet so there's no ETA yet. I'll try to look into it between other critical and high priority tasks since I'm genuinely interested in finding the cause but the impact of this particular issue is relatively low. You can also play with different custom configurations in the meanwhile to try to reproduce this behavior and get more understanding on it if you have some free time

KarboDuck:
Maybe it's just a case of new json file with smaller list of files?

If you launch db1000n with previous json (from github history) net load is still high.

`https://raw.githubusercontent.com/db1000n-coordinators/LoadTestConfig/0287df8617728449359bf0f71fa7d59fc4d627ed/config.json`

dmy3tka:
> Maybe it's just a case of new json file with smaller list of sites?
> 
> 
> 
> If you launch db1000n with previous json (from github history) net load is still high.
> 
> 
> 
> `./db1000n -c https://raw.githubusercontent.com/db1000n-coordinators/LoadTestConfig/0287df8617728449359bf0f71fa7d59fc4d627ed/config.json`

Can you please give approximate numbers on how many targets was, and is now? I hardly believe it can go down from 1 container consumes 80% cpu core to < 10% cpu core because of a smaller target list.

[...]

PR #161, opened 2022-03-05
fixed async

industral:
fixed https://github.com/Arriven/db1000n/issues/133


jdoe7865623:
You're basically making async non-configurable. Should this be fixed in configs instead?

dmy3tka:
@jdoe7865623 is right here. Can you check if we don;t have any leakage in go routine `go sendRequest()` itself. I create MR in config to add async where it was.

Arriven:
I'll close this PR as async is meant to be configurable (see my comment on the linked issue)

PR #163, opened 2022-03-05
Remove built-in VPN tunnel

bitshape:
I'm removing code I've introduced in https://github.com/Arriven/db1000n/pull/74 as I think that using `docker-compose` is more elegant solution and it's easier to work with.

@kraftwerk28 suggested to use `dperson/openvpn-client` for OpenVPN container in their PR https://github.com/Arriven/db1000n/pull/144/files. I've done some investigation and I think using https://github.com/wfg/docker-openvpn-client is better mainly due to the fact that it doesn't enforce any hard rules on how your certificates should be structured: 

https://github.com/dperson/openvpn-client/blob/a08d70a5fd4c192390305accd5593f3348e04d4d/openvpn.sh#L306

I've also opened a PR for `docker-openvpn-client` to fix an issue with `up/down` commands not being present in the `.conf` file: https://github.com/wfg/docker-openvpn-client/pull/50

I'm also adding a more complete `docker-compose.yaml` example which:
* Establishes three VPN tunnels
* Two VPNs go to one provider to two different endpoints
* One VPN goes to another provider
* Three `db1000n` containers are created, each using it's own tunnel


Arriven (approved):
LGTM but need to fix merge conflicts

docs/advanced-users-and-devs.md
```diff
@@ -50,33 +50,7 @@ Make sure you've set all available resources to docker:
...
-- Find the line in your `openvpn-provider-1.conf` where it says `auth-user-pass` and replace with `auth-user-pass provider1.txt`
-- Create a text file in `openvpn/provider1.txt` with two lines:
-
-```
-<your provider 1 username>
-<your provider 1 password>
-```
-
-- Repeat steps above for multiple providers
+See [docker-vpn](docker-vpn.md) for instructions on setting it up
```
Amet13, line 53:
```suggestion
See [docker-vpn.md](docker-vpn.md) for instructions on setting it up
```

PR #168, opened 2022-03-06
Heroku support

Amet13:
- add heroku support
- add billing warning
- fix kubernetes setup


terraform/heroku/terraform.tfvars
```diff
@@ -0,0 +1,7 @@
+# Latest release: https://github.com/Arriven/db1000n/releases/latest
+app_version = "0.5.13"
```
Amet13, line 2: It will work from the next release, as Procfile will be included to the release

Issue #146, opened 2022-03-05
gcp-installation-ua.md  not working 


sergnochevny:
cloud instance becomes unavailable after running openvpn3.sh from gcp-installation-ua.md
checked on DO and GCP

Amet13:
@Arriven @iamtodor as I mentioned before in this comment: https://github.com/Arriven/db1000n/issues/118#issuecomment-1059743438 this doc looks a bit strange. Can we find an author of it to review doc?
Or we can delete it to not to confuse people with it

iamtodor:
@Amet13 @Arriven as I suggested before let's keep only English version

Arriven:
Let's find the author first, if we get no feedback in 12 hours we can delete it but there's no rush yet since these guides expect at least some level of technical expertise which might be enough to get through even with errors in the guide

Arriven:
I think we should also enforce the rule where at least two other people should validate new guides/big changes to guides and different scripts/terraform/k8s_manifests/docker-compose files which we don't have CI jobs for yet. I'll create a separate label for that

PR #169, opened 2022-03-06
Remove gcp-ua doc cause it's not working

Amet13:
As we didn't find original author of doc, removing it

Closes https://github.com/Arriven/db1000n/issues/146


Amet13:
@Arriven @iamtodor as discussed here https://github.com/Arriven/db1000n/issues/146#issuecomment-1059794244 pingin both of you for this type of changes

PR #173, opened 2022-03-06
Slowloris - handle config reload

AndriiChuzhynov:
Use stopChan to stop workers. 
Send number "stop" signals according to configured workers

### DIFF
--- a/src/jobs/slowloris.go
+++ b/src/jobs/slowloris.go
@@ -46,12 +46,12 @@ func slowLorisJob(ctx context.Context, args Args, debug bool) error {
 	shouldStop := make(chan bool)
 	go func() {
 		<-ctx.Done()
-		shouldStop <- true
+		close(shouldStop)
 	}()
 
 	if debug {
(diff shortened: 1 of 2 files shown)


src/jobs/slowloris.go
```diff
@@ -48,9 +48,11 @@ func slowLorisJob(ctx context.Context, l *logs.Logger, args Args) error {
 	shouldStop := make(chan bool)
 	go func() {
 		<-ctx.Done()
-		shouldStop <- true
+		for i := 0; i < jobConfig.DialWorkersCount; i++ {
+			shouldStop <- true
```
Arriven, line 52:
```suggestion
			close(shouldStop)
```

Arriven (approved):
Use close to notify all workers at once (no need for the loop). This is what context package does

src/jobs/slowloris.go
AndriiChuzhynov: Yes, it's a really good point!

PR #176, opened 2022-03-06
Update cloud defaults to point to advanced image

### LINKED ISSUE #175: Change cloud instructions to use db1000n-advanced by default
I've created a separate image pointing to another config instance using specific attacks that are harder to detect and don't impact bills as much. We should switch guides for clouds to use it by default to avoid issues like high traffic billing or cloud provider locking up the instance thinking it was compromised

Arriven:
Fix #175

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,219 @@
+# Death by 1000 needles
+
+Developed by Arriven (https://github.com/Arriven)
+
+## Інструкція для новачків як встановити на виділений сервер google clouds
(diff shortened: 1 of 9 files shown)


docs/README.md
```diff
@@ -0,0 +1,219 @@
+# Death by 1000 needles
```
Amet13, line 1: @Arriven is that deployed by accident?
AFAIR we removed this file some time ago

PR #178, opened 2022-03-06
Add FAQ

### LINKED ISSUE #170: DDoS effectiveness is not directly tied to the amount of traffic generated
The app is configurable to generate set amount of traffic (controlled by the number of targets, their type, and attack interval for each of them). The main reason it works that way is because there are two main types of ddos:

- Straightforward load generation (easy to implement, easy to defend from) - as effective as the amount of raw traffic you can generate

- Actual denial of service that aims to remain as undetected as possible by simulating plausible traffic and hitting concrete vulnerabilities in the target (or targets). This type of ddos doesn't require a lot of traffic and thus is mostly limited by the amount of clients generating this type of load (or rather unique IPs)

A lot of people are worried about this app generating low amount of traffic so we need to add a section to the docs explaining this reasoning to prevent questions

Amet13:
Closes #170 

@Arriven @iamtodor let me know if we can add more info here


iamtodor:
@Amet13 as for now it seems good for me. let's fix the checks

iamtodor:
@Amet13 altho, I advise leaving the link to FAQ.md in main README.md not just having standalone FAQ

.github/workflows/test-installer.yaml
```diff
@@ -2,6 +2,8 @@ on:
   push:
     branches:
       - "main"
+    paths:
```
Amet13, line 5: Let's run installer tests only if file changed

FAQ.md
```diff
@@ -0,0 +1,39 @@
...
+- Straightforward load generation (easy to implement, easy to defend from) - as effective as the amount of raw traffic you can generate
+
+- Actual denial of service that aims to remain as undetected as possible by simulating plausible traffic and hitting concrete vulnerabilities in the target (or targets). This type of ddos doesn't require a lot of traffic and thus is mostly limited by the amount of clients generating this type of load (or rather unique IPs)
+
+## Should I care about costs if I run an app in public cloud?
+
+TL;DR: [yes](https://github.com/Arriven/db1000n/issues/153)
+
+Cloud providers could charge a huge amount of money not only for compute resources but for traffic as well.
+If you run an app in the cloud please control your billing
```
Arriven, line 39: Add reference to `ghcr.io/arriven/db1000n-advanced` as a safer option for billing

Arriven (approved):
LGTM, left minor suggestions that can be fixed later

PR #181, opened 2022-03-06
Setup for github pages with vpn instructions.

Devaniti:
@Arriven you'll need to enable github pages for it to work
You should be able to do that here https://github.com/Arriven/db1000n/settings/pages
Please select docs/ as root for pages when enabling
You can also check how will it look here https://devaniti.github.io/db1000n/system-pages/vpn


docs/system-pages/vpn.md
```diff
@@ -0,0 +1,29 @@
+# Налаштування VPN / VPN Setup
+
+## Українська
+
+Якщо ви бачите цю сторінку це означає що у вас Українська IP адресса
```
su-hs, line 5: адресса => адреса

Devaniti:
```suggestion
Якщо ви бачите цю сторінку це означає що у вас Українська IP адреса  
```

docs/system-pages/vpn.md
```diff
@@ -0,0 +1,29 @@
+# Налаштування VPN / VPN Setup
```
Amet13, line 1: docs in Ukrainian should be with `-ua` suffix

Devaniti: It have both Ukrainian and English.
We need to have both on same page, since if you forget to enable VPN, this page will be opened.
It also not supposed to be opened manually.

Amet13: I see. LGTM then.
Maybe @Arriven @iamtodor have concerns

iamtodor: LGTM +1

iamtodor (approved):

PR #184, opened 2022-03-06
add few question to faq

iamtodor:


FAQ.md
```diff
@@ -37,3 +37,30 @@ TL;DR: [yes](https://github.com/Arriven/db1000n/issues/153)
 
 Cloud providers could charge a huge amount of money not only for compute resources but for traffic as well.
 If you run an app in the cloud please control your billing (if you use Docker, ensure that use advanced image: `ghcr.io/arriven/db1000n-advanced`)
+
```
Amet13, line 40: Should we also mention in the FAQ where do we keep [hosts](https://github.com/db1000n-coordinators/LoadTestConfig) for "load testing"?

iamtodor: @Amet13 it might be. I didn't have the intention to include this information in my PR. Perhaps, you could add this info to the next one :)

PR #207, opened 2022-03-07
Added global http proxy argument for using with tor in k8s or in dock…

lollypot:
Now you can set -http-proxy 172.17.0.1:8080 as argument and use tor for http runner

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,6 +50,7 @@ func main() {
 	var configFormat string
 	var prometheusPushGateways string
 	var prometheusOn bool
+	var httpProxyUrl string
 	flag.StringVar(&configPaths, "c", "https://raw.githubusercontent.com/db1000n-coordinators/LoadTestConfig/main/config.json", "path to config files, separated by a comma, each path can be a web endpoint")
 	flag.StringVar(&backupConfig, "b", config.DefaultConfig, "raw backup config in case the primary one is unavailable")
 	flag.DurationVar(&refreshTimeout, "r", time.Minute, "refresh timeout for updating the config")
@@ -60,6 +61,7 @@ func main() {
 	flag.StringVar(&configFormat, "format", "json", "config format")
 	flag.BoolVar(&prometheusOn, "prometheus_on", false, "Start metrics exporting via HTTP and pushing to gateways (specified via <prometheus_gateways>)")
 	flag.StringVar(&prometheusPushGateways, "prometheus_gateways", "", "Comma separated list of prometheus push gateways")
+	flag.StringVar(&httpProxyUrl, "http-proxy", "", "http proxy url, override client.proxy_urls in jobs config")
 	flag.Parse()
 
 	if help {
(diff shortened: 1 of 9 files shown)


lollypot:
@Arriven reworked in more universal way

src/utils/countrychecker.go
```diff
@@ -46,7 +46,7 @@ func CheckCountry(countriesToAvoid []string) {
 
 	for _, country := range countriesToAvoid {
 		if ipInfo.Country == country {
-			log.Printf("Current country: %s. You might need to enable VPN.", ipInfo.Country)
+			log.Printf("Current country: %s. You might need to enable VPN or proxy.", ipInfo.Country)
```
Devaniti, line 49: Does country checker use proxy when doing api request?
Does it correctly report proxy country?
There are also attacks that can't pass through http proxy, so not sure how to communicate that message better.

lollypot: > Does country checker use proxy when doing api request? Does it correctly report proxy country? There are also attacks that can't pass through http proxy, so not sure how to communicate that message better.

I am working on socks5 proxy too. And no the `country checker` don't use proxy - it's just info message. But its same for vpn I guess no difference to change it for now.

Arriven: @lollypot I wondered if it would be easier to use system proxy settings here and prepared a simple [PR](https://github.com/Arriven/db1000n/pull/242) to do so. I assume that for your use case it would be enough (or even better) but not completely sure this would handle socks5 proxy that easy but we can have a separate discussion about that

Arriven: for socks5 proxy we could probably use something like this
https://github.com/valyala/fasthttp/blob/59f94a3f7137d0ad215bd262df7082d1e632a1f7/fasthttpproxy/socks5.go

lollypot: @Arriven Please take a look this https://github.com/lollypot/k8s-runners/tree/master/helm/tor-proxy and this https://github.com/lollypot/k8s-runners/tree/master/helm/runners. I just need to have possibility to set http and socks5 proxy for tor app in container and run tor in many exemplars with k8s service lb and many exemplars of db1000n. And look here https://metrics.torproject.org/rs.html#search/country:ru 	**Tota	766.55 MiB/s**. Solution with vpn or system/worker node/host proxy require much more configuration and support of k8s environment. Can you help me with this?

Arriven: I meant would it work for you to do something of the following instead of using separate commandline argument?
`http_proxy=http://proxy-server.mycorp.com:3128 ./db1000n`
Only looking in that direction because it's a more standardized way of doing things

Arriven: @lollypot check out #253 - I used your PR as a base but this should allow using any kind of proxy including socks5 ones. Would be great if you could test that one out today with your setup

lollypot: Both variants are work very well, thank you! But it seems we have other types of jobs which require socks5 proxy, especially packetgenJob. As I see in go library we can use something like this https://cs.opensource.google/go/x/net/+/27dd8689:proxy/proxy.go;l=40 and set it like `docker run -e HTTP_PROXY=http://172.17.0.1:8080 -e ALL_PROXY=soks5://172.17.0.1:9150` I have tried to set it(`ALL_PROXY`) but it not works for packetgen. Can you say me how I can implement this dealer for this job or just add support of ALL_PROXY system variable with schemes? I am not so experienced with go-lang yet.

I propose to move conversation here: https://github.com/Arriven/db1000n/issues/279

Issue #225, opened 2022-03-07
eats up 2 GB of RAM in ~2 mins and crashes with OOM error


nonchalant-enthusiast:
behavior is unstable on 2GB VPS, Ubuntu 20.04 using latest docker v0.15.8. while generating approx. the same amount of traffic

![Screenshot (31)](https://user-images.githubusercontent.com/27624986/157137074-935b5e6f-aa0c-4de9-a431-9744b6a78119.png)

![Screenshot (32)](https://user-images.githubusercontent.com/27624986/157137105-dcf85aea-6ec1-43b0-a72e-58abc9a5a3e5.png)
each spike corresponds to the deployment and consequent crash

v0.15.4 seems to be stable

sivillakonski:
@nonchalant-enthusiast , thanks for submitting the issue. I didn't use the profiler this time, so there might be other memory leaks around. Please report if you keep observing any memory related issues.

By the way, what tool do you use to see the memory consumption in this fancy way? Is it Cockpit?

Arriven:
I think the issue might be deeper than the one addressed in the referenced PR. I also don't experience it running latest code as a regular executable but do experience it in docker. Looking at it

Arriven:
ok, I've found what it is, poked the author and fixing now

PR #226, opened 2022-03-08
Fix memory leak for the non-closed response body

sivillakonski:
Addressing the issue: https://github.com/Arriven/db1000n/issues/225
References:
* http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#close_http_resp_body

Stay safe, take care, and benchmark only appropriate resources (that you own).

### DIFF
--- a/src/jobs/http.go
+++ b/src/jobs/http.go
@@ -180,6 +180,10 @@ func sendRequest(client *http.Client, req *http.Request, debug bool) {
 	}
 
 	resp, err := client.Do(req)
+	if resp != nil {
+		resp.Body.Close() // No need for response
+	}
+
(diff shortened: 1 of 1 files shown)


Arriven:
While this does seem like a good change I'm not convinced it would fix a memory leak (the only way it would be a leak is if metrics were to panic but that should've been visible on logs. There should be something else happening here, digging deeper for now

Arriven:
The original memory issue was completely different but I've also implemented alternative http job that uses fasthttp and is a lot more effective. For now it's possible to use both of them but I plan to replace default http job with it once I test it enough

PR #228, opened 2022-03-08
Avoid pushing metrics with empty string for push gateways

Lagovas:
# Description

`""` value splitted by delimiter returns `[""]` value and miss `len(var) == 0` check. So added extra check to handle this case.

## Type of change

- [+] Bug fix (non-breaking change which fixes an issue)


src/metrics/prometheus.go
```diff
@@ -163,7 +163,8 @@ func ExportPrometheusMetrics(ctx context.Context, gateways string) {
 }
 
 func pushMetrics(ctx context.Context, gateways []string) {
-	if len(gateways) == 0 {
+	// case of empty string splitted by empty delimiter
+	if len(gateways) == 1 && gateways[0] == "" {
```
Arriven, line 167: wouldn't it be safer to check for empty strings when pushing to each gateway? I know we have a vet code at the start of the program but my experience says that such dependencies don't tend to hold for long

Lagovas: we choose one gateway randomly below and use it until it failed on push. So if we will have `[""]` and check `gateway != ""`, then will be chosen the same empty gateway in the next iteration. So better to validate input data and then rely on verified correct values. imho

Arriven: Wouldn't it be easier to check for empty string in ExportPrometheusMetrics before calling this function? It's private so nobody would call it otherwise anyway and it makes a better and more intuitive check overall

Arriven:
Fixed in https://github.com/Arriven/db1000n/commit/737bf6b891068a799de23b6f8d3d770e61ad3ff5

src/metrics/prometheus.go
Lagovas: agree. changed

PR #256, opened 2022-03-09
Remove host regex matching in packetgen; add cpu profiling option

jdoe7865623:
This generates about 1.5x more traffic with the current config


Arriven:
Can you split this into two separate PRs? host regex matching change seems like a no-brainer to accept but I have some suggestions about the CPU profile option and wouldn't want them to block that regex change from being accepted

PR #262, opened 2022-03-09
Golangci lint

Arriven:
# Description

Add golangci-lint action to automate linting and catch issues early

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Test Configuration

- Release version:
- Platform:

## Logs

```text
logs
```

## Screenshots

- [ ] No screenshot

### DIFF
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -27,7 +27,7 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v3.1.0
+          version: latest
 
           # Optional: working directory, useful for monorepos
           # working-directory: somedir
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,0 +1,50 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
+  pull_request:
+    paths:
+      - "src/**"
+      - "main.go"
+      - "go.mod"
+      - "go.sum"
+      - ".github/workflows/golangci-lint.yaml"
+permissions:
+  contents: read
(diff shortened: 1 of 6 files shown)


.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,50 @@
...
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  # pull-requests: read
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-go@v2
```
vtopc, line 25: this is kinda redundant if `golangci/golangci-lint-action@v2` is used.

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.29
```
vtopc, line 31: This is a very old version. Why not use newer?

.github/workflows/golangci-lint.yaml
Arriven: took it directly from their example but let's see if it can work without it (I'd be glad to drop it if it works)

src/metrics/prometheus.go
```diff
@@ -158,7 +158,10 @@ func ExportPrometheusMetrics(ctx context.Context, gateways string) {
 	}
 	go func(ctx context.Context, server *http.Server) {
 		<-ctx.Done()
-		server.Shutdown(ctx)
+		err := server.Shutdown(ctx)
+		if err != nil {
+			log.Println("failure shutting down prometheus server:", err)
```
vtopc, line 163: The `http.ErrServerClosed`, that might be returned by the `server.Shutdown()` is not a failure.

Arriven: good point

PR #237, opened 2022-03-08
Digital Ocean Terraform code to create multiple instances in multiple…

dddbbbsss:
This module creates a number of servers in each of provided region. If you set `count = 2` and regions = ["nyc1", "nyc2", "nyc3"] this will create six servers total. Two servers in each of regions. See readme.txt for more details.

PR #263, opened 2022-03-09
Digital Ocean Terraform code to create multiple instances in multiple…

dddbbbsss:
Add version support


Arriven:
This PR is the same as #237 
Please update that one instead

PR #277, opened 2022-03-10
Add complex job scenarios

Arriven:
# Description

Adding jobs that would allow to create complex attack patterns

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Test Configuration

- Release version:
- Platform:

## Logs

```text
logs
```

## Screenshots

- [ ] No screenshot

### DIFF
--- a/testconfig.yaml
+++ b/testconfig.yaml
@@ -1,15 +1,17 @@
 jobs:
-  - type: http
+  - type: sequence
     args:
-      method: GET
-      path: "https://localhost:8080/search?searchid={{ random_uuid }}&l10n=ru&reqenc=&text={{ random_uuid }}"
-      client:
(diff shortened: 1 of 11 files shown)


testconfig.yaml
```diff
@@ -1,15 +1,17 @@
 jobs:
-  - type: http
+  - type: sequence
```
Lagovas, line 2: why don't use parallel as the default example?) it's more effective for ddos, especially in golang that well works with goroutines and blocking operations and better utilize cpu

Arriven: I'm working on the functionality, parallel is just there as a cherry on the pie, need to make sequence work with logic first

PR #297, opened 2022-03-11
add httpattack client

bmirnoff:
# Description

This is refactoring of HTTP job that causes problematic enhancement of http attacking logic.

Main concept: more HTTP attack logic into separate go files. The job only allocates HttpAttackClient and executes required methods.

This refactoring will help to add new options for HTTP attacking (like Cloudflare bypass) 

# Issue 

#298

### DIFF
--- a/src/httpattack/httpattack.go
+++ b/src/httpattack/httpattack.go
@@ -1,0 +1,273 @@
+package httpattack
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"log"
+	"math/rand"
+	"net"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/corpix/uarand"
+	"github.com/google/uuid"
+	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/fasthttpproxy"
+
+	"github.com/Arriven/db1000n/src/metrics"
+	"github.com/Arriven/db1000n/src/utils"
+	"github.com/Arriven/db1000n/src/utils/templates"
+)
+
+func NewHttpAttackClient(clientCfg map[string]interface{}, proxyURL string, debug bool) (client *HttpAttackClient) {
+	var clientConfig struct {
+		TLSClientConfig *tls.Config    `mapstructure:"tls_config,omitempty"`
+		Timeout         *time.Duration `mapstructure:"timeout"`
+		ReadTimeout     *time.Duration `mapstructure:"read_timeout"`
+		WriteTimeout    *time.Duration `mapstructure:"write_timeout"`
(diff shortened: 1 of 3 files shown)


src/httpattack/httpattack.go
```diff
@@ -0,0 +1,273 @@
...
+			log.Printf("Error sending request %v: %v", req, err)
+		}
+	}
+
+	return dataSize, err
+}
+
+func (c *HttpAttackClient) SendSingleRequest(ctx context.Context,
+	method, path, body string,
+	headers map[string]string, cookies map[string]string, debug bool) (int, *fasthttp.Response, error) {
```
Arriven, line 167: it is unsafe to return fasthttp.Response after it has been released

src/httpattack/httpattack.go
```diff
@@ -0,0 +1,273 @@
...
+	"github.com/google/uuid"
+	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/fasthttpproxy"
+
+	"github.com/Arriven/db1000n/src/metrics"
+	"github.com/Arriven/db1000n/src/utils"
+	"github.com/Arriven/db1000n/src/utils/templates"
+)
+
+func NewHttpAttackClient(clientCfg map[string]interface{}, proxyURL string, debug bool) (client *HttpAttackClient) {
```
Arriven, line 24: I do like the idea of the change but it's better to expose ClientConfig interface to the world and let the job handle parsing (see how it's done in packetgen)

src/httpattack/httpattack.go
```diff
@@ -0,0 +1,273 @@
...
+type HttpAttackClient struct {
+	client *fasthttp.Client
+}
+
+// Attacking provided path. Return transfered data
+func (c *HttpAttackClient) Attack(ctx context.Context,
+	method, path, body string,
+	headers map[string]string, cookies map[string]string, debug bool) (int, error) {
+
+	methodTpl, pathTpl, bodyTpl, headerTpls, cookieTpls, err := prepareHTTPRequestTemplates(method, path, body, headers, cookies)
```
Arriven, line 125: the point of splitting parsing and execution of templates was performance - parsing is quite expensive and shouldn't be done in a loop. Here you essentially break that idea completely by moving parsing back into the loop iteration

Arriven (commented):
While the idea of moving http related functionality into a separate module is good this implementation introduces several critical correctness and performance degradations so I can't accept this PR in its current state. In the issue you've mentioned that you want to work on bypassing cloudflare but from what I can tell it will require a separate type of job to parse and execute javascript while http requests logic shouldn't change that much (it already allows bypassing some of ddos-guard defences with proper configuration). I'd like to see some kind of technical write-up on what you want to achieve before you start working on next steps in this direction

PR #317, opened 2022-03-13
DB1000N | Add app. auto-restart upon successful update

sivillakonski:
# Over-the-air updates

Lots of maintainers run their needles on a bare metal machines.
As long as this project is so frequently updated, it might be
a good idea to let them update it without the hassle.

- [V] Enabled automatic time-based version check
- [V] Enabled application self-restart after it downloaded the update

## Description

Support for the application self-update by downloading
the latest release from the official repository.

```text
Stay strong, be the first in line!
```

The version should be embedded in the binary during the build, see the `build`
target in the Makefile.

### Available flags

```bash
  -enable-self-update
        Enable the application automatic updates on the startup
  -restart-on-update
        Allows application to restart upon the successful update (ignored if auto-update is disabled) (default true)
  -self-update-check-frequency duration
        How often to run auto-update checks (default 24h0m0s)
  -skip-update-check-on-start
        Allows to skip the update check at the startup (usually set automatically by the previous version) (default false)
```

The default behavior if the self-update enabled:
1. Check for the update
2. If update is available - download it
   1. If auto-restart is enabled
      1. Notify the user that a newer version is available
      2. Fork-Exec a new process (will have a different PID), add a flag to skip the version check upon startup
      3. Stop the current process
   2. If auto-restart is disabled - notify user that manual restart is required
3. If update is NOT available - schedule the next check

### Examples

To update your needle, start it with a flag `-enable-self-update`

```sh
./db1000n -enable-self-update
```

#### Advanced options

* Start the needle with the **self-update & self-restart**
```bash
$ ./db1000n -enable-self-update 
0000/00/00 00:00:00 main.go:82: DB1000n [Version: v0.6.4][PID=75259]
0000/00/00 00:00:00 main.go:166: Running a check for a newer version...
0000/00/00 00:00:00 main.go:176: Newer version of the application is found [0.7.0]
0000/00/00 00:00:00 main.go:177: What's new:
* Added some great improvements
* Added some spectacular bugs
0000/00/00 00:00:00 main.go:180: Auto restart is enabled, restarting the application to run a new version
0000/00/00 00:00:00 restart.go:45: new process has been started successfully [old_pid=75259,new_pid=75262]

# NOTE: Process 75259 exited, Process 75262 has started with a flag to skip version check on the startup

0000/00/00 00:00:00 main.go:82: DB1000n [Version: v0.7.0][PID=75262]                                                                             
0000/00/00 00:00:00 main.go:155: Version update on startup is skipped, next update check is scheduled in 24h0m0s
```

* Start the needle with the self-update but do not restart the process upon update (`systemd` friendly)
```bash
$ ./db1000n -enable-self-update -self-update-check-frequency=5m -restart-on-update=false
0000/00/00 00:00:00 main.go:82: DB1000n [Version: v0.6.4][PID=75320]
0000/00/00 00:00:00 main.go:166: Running a check for a newer version...
0000/00/00 00:00:00 main.go:176: Newer version of the application is found [0.7.0]
0000/00/00 00:00:00 main.go:177: What's new:
* Added some great improvements
* Added some spectacular bugs
0000/00/00 00:00:00 main.go:191: Auto restart is disabled, restart the application manually to apply changes!
```

# References

* https://github.com/Scalingo/go-graceful-restart-example
* https://github.com/rcrowley/goagain
* https://grisha.org/blog/2014/06/03/graceful-restart-in-golang


sivillakonski:
@Arriven, please, add the version embedding to the binary distributed thru release downloads. The example on how to add the version you can find in the `Makefile`.

Btw, I can't test if the restart works fine for the Windows machines, I don't have one.

main.go
```diff
@@ -119,13 +126,63 @@ func main() {
...
+
+func watchUpdates(doRestartOnUpdate, skipUpdateCheckOnStart bool, autoUpdateCheckFrequency time.Duration) {
+	if !skipUpdateCheckOnStart {
+		runUpdate(doRestartOnUpdate)
+	} else {
+		log.Printf("Version update on startup is skipped, next update check is scheduled in %s",
+			autoUpdateCheckFrequency)
+	}
+
+	periodicalUpdateChecker := time.NewTicker(autoUpdateCheckFrequency)
```
Arriven, line 152: Even though this function only returns on exiting the program, let's still defer ticker.Stop() call here

src/utils/ota/ota.go
```diff
@@ -2,80 +2,41 @@
...
 
-// DoSelfUpdate updates the app to the latest version
-func DoSelfUpdate() {
+// DoAutoUpdate updates the app to the latest version.
+func DoAutoUpdate() (updateFound bool, newVersion, changeLog string, err error) {
 	v, err := semver.ParseTolerant(Version)
 	if err != nil {
-		log.Println("Binary version validation failed:", err)
-
+		err = errors.Wrap(err, "binary version validation failed")
```
Arriven, line 14: Please use fmt.Errorf here, no need to import additional dependency just to wrap errors (even if it's faster - this is not a hot path)

src/utils/ota/restart.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	log.Printf("new process has been started successfully [old_pid=%d,new_pid=%d]\n",
+		os.Getpid(), fork)
+
+	os.Exit(0)
+	return nil
+}
+
+func mergeExtraArgs(osArgs, extraArgs []string) []string {
```
Arriven, line 52: This function is unnecessary complex and doesn't save us from errors: what if extraArgs is something like ["--pprof",":8081"] and os args already include --pprof :8080 ?

Arriven (commented):
Looks like a promising change, but some polishing needed

src/utils/ota/ota.go
```diff
@@ -2,80 +2,41 @@
 package ota
 
 import (
-	"bufio"
-	"fmt"
-	"log"
-	"os"
-
 	"github.com/blang/semver"
+	"github.com/pkg/errors"
```
vtopc, line 6: This package is deprecated.

sivillakonski: That's interesting, didn't know about it.

src/utils/ota/restart.go
sivillakonski: I would be careful not accepting duplicated args here, because there are programs accepting multiple flags. For example - docker :) You can add multiple ports with the `-p` flag: `-p 8080:80/tcp -p 8080:80/udp`

sivillakonski: The only purpose of passing additional arguments to the next process is allow it to skip the version check at the startup with the `-skip-update-check-on-start`. It's not necessary, but I was aware of the situation where there is still no version is being embedded into the binary file in the release, so this way it would be an endless update loop.

~~~
(1) Binary version v0.0.1 (default) -> (2) Finds a version v0.7.0 (no version is embedded to it) -> (3) v0.0.1 Shuts down to run the v0.7.0 -> (4) v0.7.0 starts without knowing what version it is -> (5) Finds v0.7.0 -> Go to step (3)
~~~

This flag allows to defer this situation, and with a delay 24 hours it would still work without the embedded version and still downloading the latest version

Arriven: I'm ok with it checking for duplicates, it's just that it should be done correctly. Also considering how current flags implementation works I'd rather move this kind of fix into another PR since other parts of it are quite straightforward but with this one we could explore multiple options and I don't want this process to prevent other changes from being merged

sivillakonski:
Checked the build with a cross-compilation in the docker container:
~~~
root@0478a04ad044:/go/src/github.com/Arriven/db1000n# go version
go version go1.16.15 linux/amd64
root@0478a04ad044:/go/src/github.com/Arriven/db1000n# GOOS=windows CGO_ENABLED=0 go build -ldflags="-s -w -X 'github.com/Arriven/db1000n/src/utils/ota.Version=v0.7.0'" -o db1000n -a ./main.go
root@0478a04ad044:/go/src/github.com/Arriven/db1000n# GOOS=darwin CGO_ENABLED=0 go build -ldflags="-s -w -X 'github.com/Arriven/db1000n/src/utils/ota.Version=v0.7.0'" -o db1000n -a ./main.go
root@0478a04ad044:/go/src/github.com/Arriven/db1000n# GOOS=linux CGO_ENABLED=0 go build -ldflags="-s -w -X 'github.com/Arriven/db1000n/src/utils/ota.Version=v0.7.0'" -o db1000n -a ./main.go
~~~

src/utils/ota/restart.go
sivillakonski: Yes, I do completely agree with you. It's much more complicated process to properly merge additional arguments, but for the given use-case, I believe, it should be fine not to duplicate a simple flag

Arriven (approved):

PR #325, opened 2022-03-13
fix: release script fixed to add correct build flags

### LINKED ISSUE #322: v0.7.0 says  DB1000n [Version: v0.0.1]
I assume it's left over hardcoded version value or somewhere the code has not been changed properly

Screenshot 2022-03-13 at 13 14 20

Reproduced via executables on MacOS
- for Intel CPU https://github.com/Arriven/db1000n/releases/download/v0.7.0/db1000n-v0.7.0-darwin-amd64.tar.gz
- for M1 CPU https://github.com/Arriven/db1000n/releases/download/v0.7.0/db1000n-v0.7.0-darwin-arm64.tar.gz

cc/ @Arriven ; @bitshape

arfgdev:
# Description

release binaries github action fixed
Fixes #322

## Type of change

### DIFF
--- a/src/utils/ota/version.go
+++ b/src/utils/ota/version.go
@@ -2,7 +2,7 @@ package ota
 
 var (
 	// Version is a releae version embedded into the app
-	Version = "v0.0.1"
+	Version = "development"
 	// Repository to check for updates
 	Repository = "Arriven/db1000n" // Could be changed via the ldflags
 )
--- a/.github/workflows/release-binaries.yaml
+++ b/.github/workflows/release-binaries.yaml
@@ -23,3 +23,4 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           goos: ${{ matrix.goos }}
           goarch: ${{ matrix.goarch }}
+          build_command: make


Arriven:
Already fixed, there was a caveat of making make build output .exe in order to comply with the action (just one more point in my "why I don't like windows" list). Changing the version name to development seems like a nice idea, need to test it out though

src/utils/ota/version.go
```diff
@@ -2,7 +2,7 @@ package ota
 
 var (
 	// Version is a releae version embedded into the app
-	Version = "v0.0.1"
+	Version = "development"
```
Arriven, line 5: @arfgdev can you ensure OTA updates functionality works well with this change?

arfgdev: +

arfgdev: it does not work 😅
`Binary version validation failed: Invalid character(s) found in major number "development"`

sivillakonski: It won't work because it doesn't follow the semver semantics

Arriven: I'll close the PR then

PR #328, opened 2022-03-13
Add option to apply your own proxy list with command-line args

Staskkk:
# Description
The app was not using a proxy list from URL when it was applied via command-line argument "-proxylist-url".

The fix allows specifying proxy list url via the command-line argument (without using custom config).

Changes summary:
Add ProxyListURL to GlobalConfig
Make ProxyListURL a priority over ProxyURL
Fix clientConfig.ProxyURLs assign
Make func GetProxylist public

Motivation: Bug that proxy list does not work and similar open issue.

No new dependencies are required

Fixes # (issue)
https://github.com/Arriven/db1000n/issues/34

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- Run Http Server locally, make it return the proxy list in the following way:
[
"socks5://89.108.81.117:1080",
"socks5://37.139.243.32:1080",
"socks5://80.71.177.6:1080"
]
Put a link to this local HTTP server as an argument to the db1000n:
go run main.go -proxylist-url http://localhost:8123

## Test Configuration

- Release version: 0.7.4
- Platform: Windows 10 x64

## Logs

```text
2022/03/13 16:26:47.079005 http.go:148: Attacking [http://23.111.238.40:80]()
2022/03/13 16:26:47.079005 http.go:103: List of proxies: [socks5://89.108.81.117:1080,socks5://37.139.243.32:1080,socks5://80.71.177.6:1080]()
2022/03/13 16:26:47.079005 http.go:109: proxyURLs: [[socks5://89.108.81.117:1080]() [socks5://37.139.243.32:1080]() [socks5://80.71.177.6:1080]()]
2022/03/13 16:26:47.079504 http.go:176: GET [https://62.213.104.59:443]() started at 1647188807
2022/03/13 16:26:47.079504 http.go:148: Attacking [http://178.248.233.60:80]()
2022/03/13 16:26:47.079504 http.go:103: List of proxies: [socks5://89.108.81.117:1080,socks5://37.139.243.32:1080,socks5://80.71.177.6:1080]()
2022/03/13 16:26:47.079504 http.go:109: proxyURLs: [[socks5://89.108.81.117:1080]() [socks5://37.139.243.32:1080]() [socks5://80.71.177.6:1080]()]
```

## Screenshots

image


Arriven (commented):
While the idea is good I think we should aim towards reducing the overall amount of configuration options and make each option more expandable. The way this particular functionality should be configured is something of the following:
The client config should take a comma separated list of proxies that comes either from the commandline (extend existing `system-proxy` flag instead of providing the new one) or from a job config (where we should set it completely within the template, i.e. `{{ join (get_url_json <url>).proxies "," }}`)
If both of those are empty - use regular unix environment variables the same way it is done now
I don't think that changing the job config is in the scope of this PR but lets at least make it so that the commandline flag part is working according to this design

Staskkk:
All right, I will extend the existing system-proxy flag (not sure why the existing -proxylist-url flag is never used though, I didn't add it, it was in the code, just never used from the command-line).
What about making a flag with a proxy list file name, similar to how it's done in MHDDoS? It makes more sense for me than providing all of the proxies in the command line.

Arriven:
@Staskkk the idea with the file seems good but you can kinda already achieve that functionality via the shell. i.e. `db1000n --system-proxy $(cat proxylist.txt | tr '\n' ',')`

Staskkk:
All right, I will try to post the changes later today, thanks!

Arriven:
actually, you can already pass a comma-separated list of proxies into the `--proxy` flag (misstyped that flag in previous comment, sorry). This won't work if you want the proxylist to be dynamically updated with an interval but even that would better be achieved by letting the commandline to be templated and tweaking the config so that it's reevaluated every 10000 requests or so) - I'll probably even work on that now while the idea is fresh in my head

Staskkk:
@Arriven All right, I'll use the --proxy then and make PR for some other issue. BTW, is it always just choosing the random proxy from the list? Maybe it makes sense to use all the proxies in the separate threads (or at least add an option for that)?
Or some other optimization, like use them in order or drop some of the proxies if they become slow?

Arriven:
It's always choosing random proxy from the list, yes, and I think it's better than splitting the list between threads as that would actually make it easier for specific targets to defend - they'd have to block only a sub list of IP addresses rather than full range of them with full range of proxies
As for checking if proxies become slow/unavailable I think it's better to handle that outside of the program as it can be done more efficiently that way

Arriven:
Also see #378 for a draft of proxylist refactoring, I think I'll finish it today a bit later

Staskkk:
All right man, thanks for the explanation!

PR #332, opened 2022-03-14
Add "updater mode"

bitshape:
# Description

This is mostly useful if you're running `db1000n` with docker-compose with VPN container. In this case you're likely to receive errors or time-outs while trying to update config file due to the fact that VPN connection is already highly saturated. Typically it looks like this:

```
runner.go:154: The app has generated approximately ... bytes of traffic
runner.go:157: Of which for ... bytes we received some response from the target
config.go:55: Failed to fetch config from "https://raw.github.com/.../config.json": Get unexpected EOF
config.go:66: Could not load new config, proceeding with the last known good one
config.go:148: The config has not changed. Keep calm and carry on!
```

Updater mode is a simple loop that will try fetching config file from remote server and store it in shared Docker volume. This is achieved by changing entrypoint to `--updater-mode` and mapping `./config` as shared volume to it. If any error is encountered, this process will exit and Docker will restart it. Healthcheck is checking for existence of this file, which means:

1) db1000n containers won't start until `updater` container pulled config first
2) If there are any other errors, `autohealer` will restart it

Normal `db1000n` containers will require two configuration changes:

1) Adding `./config` directory as shared volume so they can read configuration that is pulled by `updater` container
2) Change entrypoint to load config from file
3) Adding `updater` to their `depends_on` list so that they won't start until config file is available

The end result is such that `db1000n` container traffic is routed through VPN container tunnels but config update is routed outside VPN tunnel which means it's more likely to succeed.

I had to refactor a bunch of stuff in order to remove initial code duplication in `updater`. Probably easiest to follow commit by commit. The most important change is that `Fetcher` is removed and `lastKnownConfig` is simply passed through.

## Type of change

### DIFF
--- a/src/utils/updater/updater.go
+++ b/src/utils/updater/updater.go
@@ -10,23 +10,23 @@ import (
 	"github.com/Arriven/db1000n/src/runner/config"
 )
 
-func Run(configPaths []string, backupConfig []byte) {
+func Run(destinationConfig string, configPaths []string, backupConfig []byte) {
 	lastKnownConfig := &config.RawConfig{Body: backupConfig}
 
 	for {
 		rawConfig := config.FetchRawConfig(configPaths, lastKnownConfig)
 
 		if !bytes.Equal(lastKnownConfig.Body, rawConfig.Body) {
-			file, err := os.Create("config/config.json")
+			file, err := os.Create(destinationConfig)
 			if err != nil {
-				log.Println("Unable to create config/config.json")
+				log.Printf("Unable to create %s", destinationConfig)
 
 				return
 			}
 
 			size, err := io.WriteString(file, string(rawConfig.Body))
 			if err != nil {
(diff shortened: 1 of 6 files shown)


src/utils/updater/updater.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+func Run(configPaths []string, backupConfig []byte) {
+	lastKnownConfig := &config.RawConfig{Body: backupConfig}
+
+	for {
+		rawConfig := config.FetchRawConfig(configPaths, lastKnownConfig)
+
+		if !bytes.Equal(lastKnownConfig.Body, rawConfig.Body) {
+			file, err := os.Create("config/config.json")
```
Arriven, line 20: Can we expose this path via commandline/env variable? Can be done in a separate PR

Arriven (approved):

Issue #339, opened 2022-03-14
Version 0.7.7 crashed on Windows10: need to fix


SerhiiZavalko:
**Powershell**
```
PS C:\Users\User\src\war\ddos\db1000n> .\db1000n
2022/03/14 22:16:05.301887 main.go:54: DB1000n [Version: v0.7.7][PID=236]
2022/03/14 22:16:05.301887 countrychecker.go:28: Checking IP address, attempt #0
2022/03/14 22:16:06.750456 countrychecker.go:97: Current country: United Kingdom (2001:ac8:21:f000::212e)
2022/03/14 22:16:06.752756 prometheus.go:213: Can't get system cert pool
2022/03/14 22:16:06.984585 config.go:60: Loading config from "https://raw.githubusercontent.com/db1000n-coordinators/LoadTestConfig/main/config.v0.7.json"
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x20 pc=0xaf5c23]

goroutine 40 [running]:
crypto/x509.(*CertPool).addCertFunc(...)
        /usr/local/go/src/crypto/x509/cert_pool.go:188
crypto/x509.(*CertPool).AppendCertsFromPEM(0x0, {0xc0000fcc00, 0x3d5, 0x79e19ff64140})
        /usr/local/go/src/crypto/x509/cert_pool.go:226 +0x263
github.com/Arriven/db1000n/src/utils/metrics.getTLSConfig()
        /github/workspace/src/utils/metrics/prometheus.go:227 +0x8a
github.com/Arriven/db1000n/src/utils/metrics.pushMetrics({0x12c8920, 0xc000067a80}, {0xc0000a5680, 0x3, 0xc000068900})
        /github/workspace/src/utils/metrics/prometheus.go:244 +0xeb
created by github.com/Arriven/db1000n/src/utils/metrics.ExportPrometheusMetrics
        /github/workspace/src/utils/metrics/prometheus.go:173 +0x225
```

**CMD**
```
C:\Users\User\src\war\ddos\db1000n>db1000n
2022/03/14 22:17:13.119499 main.go:54: DB1000n [Version: v0.7.7][PID=16144]
2022/03/14 22:17:13.120050 countrychecker.go:28: Checking IP address, attempt #0
2022/03/14 22:17:14.523883 countrychecker.go:97: Current country: United Kingdom (2001:ac8:21:f000::212e)
[...]

Issue #341, opened 2022-03-14
Feature request: "-use-advanced-config" flag for binary distribution


dmitry-salnikov:
Since you've added self-update for binary distribution - thanks! - could you please add the CLI option to use advanced config - same as used by docker image `ghcr.io/arriven/db1000n-advanced` mentioned in the docs.

It's preferrable to add kind of `-use-advanced-config` flag instead of specifying exact config path, because when running the tool like
```sh
./db1000n -enable-self-update -с https://raw.githubusercontent.com/db1000n-coordinators/LoadTestConfig/main/config.v0.7.adv.json
```
the binary eventually updates to 0.8, and config version 0.7 may become incompatible.

So, it would be awesome to be able to run the binary as:

```
./db1000n -enable-self-update -use-advanced-config
```

sivillakonski:
Hi folks, thank you for your feedback!
Lemme see what we can do with this :)

PR #342, opened 2022-03-15
DB1000N | Add advanced configuration default dynamic URL based on version & fix panic on missing system cert pool for the Prometheus Metrics

sivillakonski:
Hello there, my fellow friends

Addressing the issue: https://github.com/Arriven/db1000n/issues/339

Fix the Windows-related issue related to the missing system cert. pool. `SystemCertPool() (*CertPool, error)` method is not available for the Windows machines (`system root pool is not available on Windows`)

@Arriven, this will only prevent the panic but there might be some changes needed to disable a call to the cert. pool for the Windows platform.


sivillakonski:
Addressing the request: https://github.com/Arriven/db1000n/issues/341

No need to provide the extra flag to automatically follow the major & minor version changes, thanks to @[dmitry-salnikov](https://github.com/dmitry-salnikov) for the proposal!

src/runner/config/defaultconfig_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"testing"
+
+	"github.com/Arriven/db1000n/src/utils/ota"
+)
+
+//nolint: paralleltest // No need to test this in parallel :p
```
Arriven, line 11: please avoid using nolint at all costs. For now it is used only when the functionality of the app actually requires it (`InsecureSkipVerify: true` and cryptographic random)

src/utils/metrics/prometheus.go
```diff
@@ -210,7 +211,7 @@ var PushGatewayCA string
 func getTLSConfig() (*tls.Config, error) {
 	rootCAs, err := x509.SystemCertPool()
 	if err != nil {
-		log.Println("Can't get system cert pool")
+		return nil, fmt.Errorf("can't get system cert pool: %w", err)
```
Arriven, line 214: I'll probably cherry-pick this change into main since it's a bug that has to be fixed asap but I can't merge the whole PR because of other comments

src/runner/config/defaultconfig.go
```diff
@@ -15,3 +22,23 @@ func init() {
...
+	if err != nil || !isVersionEmbeddedToBinary {
+		return FallbackConfigURL
+	}
+
+	major, minor, patch, err := ota.ResolveVersion()
+	if err != nil {
+		return FallbackConfigURL
+	}
+
+	return buildDefaultConfigURLForVersion(major, minor, patch)
```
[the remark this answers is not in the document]
Arriven: ok, I see which use-case is being solved here. It would probably be wiser to use filters on jobs which I've added couple days ago instead of having two separate configs. Let me work on that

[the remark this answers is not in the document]
sivillakonski: It's completely up to you, @Arriven. Probably, this dynamic URL build is overkill for this

src/runner/config/defaultconfig_test.go
sivillakonski: So, I would rather re-configure the linter that gives the warning. This is just a unit test that validates 10 lines of code - no need to add any kind of parallel testing here

Arriven: I feel like parallel testing is used here in order to speed up the testing process

src/utils/metrics/prometheus.go
Lagovas: hm, I saw that this problem was solved in fresh go1.18 for windows )) https://go.dev/doc/go1.18#crypto/x509
And I made it incorrectly. Better create empty cert pool instead of error because there is enough to create empty pool
In another project we do next:
```
var roots *x509.CertPool
	var err error
	// use system pool as default
	if roots, err = x509.SystemCertPool(); err != nil {
		log.WithError(err).WithField(logging.FieldKeyEventCode, logging.EventCodeErrorGeneral).
			Errorln("Can't load system ca certificates")
	}
	if roots == nil {
		roots = x509.NewCertPool()
	}
```

Arriven: @Lagovas good point, also thought about that. Unfortunately I'll have to wait with switching to go1.18 as not all the tooling I use support it yet (but I'll do the empty pool creation still)

Lagovas: keep in mind that empty cert pool will works with our gates that use self-signed certificate but will fail for domain-based gates' URLs, because of need system cert pool with global known CA certs. but it should not panic, just pushing metrics will not work

Arriven: I understand, it's still better than not working metrics on windows even for IP-based gateways

src/runner/config/defaultconfig_test.go
sivillakonski: The problem here is that's unique case where a global variable is in the place (`ota.Version`) and running these test in parallel would lead to the problem of its concurrent modification.

sivillakonski:
@Arriven , mind if I closed the PR? Looks like this feature is not relevant here

Arriven:
@sivillakonski sure

PR #380, opened 2022-03-20
Enable cyclop and make it happy

jdoe7865623:


src/core/dnsblast/qry/types.go
```diff
@@ -23,148 +23,81 @@ func ResponseCode(rc int) string {
...
-		return dns.TypeEUI64
-	case "URI":
-		return dns.TypeURI
-	case "CAA":
-		return dns.TypeCAA
-	case "AVC":
-		return dns.TypeAVC
-	default:
+	res, ok := map[string]uint16{
+		"None":       dns.TypeNone,
```
Arriven, line 27: Does cyclop bitch about switch because it thinks of each case as a separate branch? Also are we sure that the map is optimized away by the compiler? (I assume it's yes for both cases but better be safe than sorry)

src/jobs/utils.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/Arriven/db1000n/src/utils/templates"
 )
 
-func logJob(ctx context.Context, logger *zap.Logger, globalConfig GlobalConfig, args Args) (data interface{}, err error) {
+func logJob(ctx context.Context, logger *zap.Logger, _ GlobalConfig, args Args) (data interface{}, err error) { //nolint:unparam // data is here to match Job
```
Arriven, line 15: I wonder why this nolint is needed here but not in other jobs that never return any data?

src/utils/templates/templates.go
```diff
@@ -25,9 +25,13 @@ func getProxylistURL() string {
 	return proxiesURL
 }
 
-// SetProxiesURL is used to override the default proxylist url
-func SetProxiesURL(url string) {
-	proxiesURL = url
+// SetProxiesURL is used to override the default proxylist url with a non-empty url.
+func SetProxiesURL(url *string) {
```
Arriven, line 29: Just FYI: this function was removed recently

Arriven (commented):
Looks good, have couple minor questions but you'll have to resolve merge conflicts

src/jobs/utils.go
jdoe7865623: Good question. I have more questions to unparam because it was enabled before this PR, but never complained here. Somehow it started complaining while I was fixing cyclop.

src/utils/templates/templates.go
jdoe7865623: Gotcha

src/core/dnsblast/qry/types.go
jdoe7865623: Yes, cyclop does not like lengthy switches. I doubt the map is optimized out and I see varying info on the strings switch vs map on the net, so I guess I'll make a perftest to see the difference.

Arriven (approved):

src/jobs/utils.go
Arriven: Guess we can tackle it separately

Issue #279, opened 2022-03-10
Support of system variable like ALL_PROXY with schemes http://ip:port and socks5://ip:port for using with tor and packetgen job type


lollypot:
Continue of https://github.com/Arriven/db1000n/pull/207

For now we can set http proxy in two variants:

1. By system variable like `HTTP_PROXY` or `http_proxy`
2. By app argument like `-proxy 172.17.0.1:8080`

Both variants are work very well. But it seems we have other types of jobs which require socks5 proxy, especially packetgenJob. As I see in go library we can use something like this https://cs.opensource.google/go/x/net/+/27dd8689:proxy/proxy.go;l=40 and https://cs.opensource.google/go/x/net/+/27dd8689:proxy/proxy.go;l=92 and set it like `docker run -e HTTP_PROXY=http://172.17.0.1:8080 -e ALL_PROXY=soks5://172.17.0.1:9150` I have tried to set it(`ALL_PROXY`) but it not work for packetgen. Can you say me how I can implement this dealer for this job or just add support of ALL_PROXY system variable with schemes? I am not so experienced with go-lang yet.
All of this need to run a bunch db1000n applications in k8s cluster with tor http and socks proxy. Take a look here https://github.com/lollypot/k8s-runners/tree/master/helm for examples.

Thank you.

Arriven:
that's actually a good question. I'd be looking at something like https://pkg.go.dev/golang.org/x/net/proxy#SOCKS5 
I'm not sure whether it would be easier to embed it into packetgen or rawnet jobs but both of them use standard go net package to send their data so it should be possible. Will take a look at it myself a bit later

Arriven:
@lollypot considering the fact that socks5 proxies don't work for udp I feel like it would also be impossible to add them to packetgen but feel free to experiment

Arriven:
Closing this as most proxies don't support udp or raw ip traffic (including tor)

PR #381, opened 2022-03-20
Added proxy URL validation and socks5 support for rawnet job type (#279)

lollypot:
See description of CR here #279

It's still not work for udp type jobs
"msg":"error running job","error":"socks connect udp 172.17.0.1:9150->1.2.3.4:53: network not implemented"
because socks5 library is not support udp for now. But I am decided to leave it as is because when you will run it through proxy - you expect that it it should go through proxy and, probably, you don't want to get abuse messages to yours workloads.

### DIFF
--- a/main.go
+++ b/main.go
@@ -117,6 +117,13 @@ func main() {
 		log.Fatal("Invalid value for --prometheus_gateways")
 	}
 
+	if *systemProxy != "" {
+		_, err = utils.ParseProxyUrl(*systemProxy)
+		if err != nil {
+			log.Fatalf("Invalid value for --proxy: %v", err)
+		}
+	}
+
 	country := ""
 
 	var isCountryAllowed bool
--- a/src/jobs/rawnet.go
+++ b/src/jobs/rawnet.go
@@ -1,6 +1,7 @@
 package jobs
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"net"
@@ -40,19 +41,20 @@ func tcpJob(ctx context.Context, logger *zap.Logger, globalConfig GlobalConfig, 
 	go processedTrafficMonitor.Update(ctx, time.Second)
 
 	for jobConfig.Next(ctx) {
-		sendTCP(ctx, logger, jobConfig, trafficMonitor, processedTrafficMonitor)
+		sendTCP(ctx, logger, globalConfig, jobConfig, trafficMonitor, processedTrafficMonitor)
 	}
 
 	return nil, nil
 }
 
-func sendTCP(ctx context.Context, logger *zap.Logger, jobConfig *rawnetConfig, trafficMonitor, processedTrafficMonitor *metrics.Writer) {
+func sendTCP(ctx context.Context, logger *zap.Logger, globalConfig GlobalConfig, jobConfig *rawnetConfig, trafficMonitor, processedTrafficMonitor *metrics.Writer) {
 	tcpAddr, err := net.ResolveTCPAddr("tcp", jobConfig.addr)
 	if err != nil {
 		return
 	}
 
-	conn, err := net.DialTCP("tcp", nil, tcpAddr)
+	conn, err := getProxyConnection(globalConfig.ProxyURLs, tcpAddr.String())
+
 	if err != nil {
 		logger.Debug("error connecting via tcp", zap.Reflect("addr", tcpAddr), zap.Error(err))
 		metrics.IncRawnetTCP(tcpAddr.String(), metrics.StatusFail)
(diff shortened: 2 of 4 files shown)


mvlm:
It's still not work for udp type jobs
`"msg":"error running job","error":"socks connect udp 172.17.0.1:9150->1.2.3.4:53: network not implemented"`
because socks5 library is not support udp for now. But I am decided to leave it as is because when you will run it through proxy - you expect that it it should go through proxy and, probably, you don't want to get abuse messages to yours workloads.

main.go
```diff
@@ -119,6 +119,13 @@ func main() {
 		log.Fatal("Invalid value for --prometheus_gateways")
 	}
 
+	if *systemProxy != "" {
+		_, err = utils.ParseProxyUrl(*systemProxy)
+		if err != nil {
+			log.Fatalf("Invalid value for --proxy: %v", err)
```
Arriven, line 125: This tool tries to follow a philosophy where it has to work to some extent regardless of how badly it's misconfigured. Considering that proxy list can be templated now it's better to avoid this type of check unless absolutely necessary (I plan to remove similar check for metrics as well). The check should be performed by the module that uses this value instead

src/utils/utils.go
```diff
@@ -123,3 +124,17 @@ func OpenBrowser(url string) {
 
 	log.Printf("Please open %s", url)
 }
+
+func ParseProxyUrl(proxyAddr string) (*url.URL, error) {
```
Arriven, line 128: Considering my previous comment about this check being used in main I think it's better to move this function straight to src/jobs/rawnet.go

Arriven (commented):
Left couple of comments, I also think it's better to not use proxy in udp jobs if it's not supported as there's now better ways to avoid using those jobs (see --enable-primitive-jobs flag and corresponding filter in testconfig.json)

Arriven:
Btw, would it be ok to you to add this proxy to slowloris as well? Can be a separate PR

Arriven:
And one more thing - can we also expose proxy settings to the job config so that it's similar to how it's done in http jobs?

Arriven:
@lollypot hope you don't mind me reimplementing this with my comments in #384

src/jobs/rawnet.go
```diff
@@ -40,19 +43,29 @@ func tcpJob(ctx context.Context, logger *zap.Logger, globalConfig GlobalConfig,
...
 
-	conn, err := net.DialTCP("tcp", nil, tcpAddr)
+	client := proxy.FromEnvironmentUsing(&net.Dialer{})
+
+	if globalConfig.ProxyURL != "" {
+		u, err := url.Parse(globalConfig.ProxyURL)
+		if err != nil {
+			return
+		}
+		client, err = proxy.FromURL(u, &net.Dialer{})
```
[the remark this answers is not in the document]
lollypot: Fixed

src/utils/utils.go
```diff
@@ -123,3 +124,17 @@ func OpenBrowser(url string) {
...
+	if proxyAddr != "" {
+		u, err := url.Parse(proxyAddr)
```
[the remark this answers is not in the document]
lollypot: Done

main.go
lollypot: Yes, I see, but I think it's very bad when you have set proxy url and its not valid and error will be ignored even if you get some message on screen but you can not see it in time. In this case you will get a buch of outgoing traffic from your pods/container without using proxy. It's can compomate you account and you will get suspended account in cloud or light up your real ip adresses. So my opinoin if you set proxyurl but if it's have some vrong format, app should not send traffic. It may not related to the jobs config but for system variable or argument value - proxy with error handling should be I think.

src/utils/utils.go
lollypot: It's was used in main.go for global proxy url validation and related not only to the rawnet job.

lollypot:
@Arriven Yes. Thank you very much - you will do it better. of course :). Sorry that I not respond in shorter time. But I have one comment: https://github.com/Arriven/db1000n/pull/381#discussion_r830877783 . What do you think?

main.go
Arriven: @lollypot we can probably work on separate commandline switch for that but I feel like this level of concern is close to a point where it's more viable to utilize your own config or to secure the vm/container you are running it in to forward all of the traffic through the proxy/vpn/tor instead (i.e. what's going to stop me from embedding a request into the template evaluation inside some other job? there's already a function for that used which I added to be able to retrieve proxylist from a custom location and it's not going through any proxy. it's highly unlikely for me to do something as stupid as using it for an attack but it's not impossible)

lollypot: Ok got it, thank you.

PR #391, opened 2022-03-22
DB1000N | Add support for the restart option after the auto-update for the Windows OS

sivillakonski:
## Restart the needle once the auto-update has finished on the Windows OS
(Codename: `Glory to Those who protect the Land`)

This patch brings the restart the option for the DB1000N on the OS with non-boring wallpapers.

Take care and stay safe, folks!

### References
1. https://github.com/golang/go/issues/29841
2. https://superuser.com/a/1656458

### DIFF
--- a/src/utils/ota/restart_windows.go
+++ b/src/utils/ota/restart_windows.go
@@ -1,0 +1,46 @@
+//go:build windows
+// +build windows
+
+package ota
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+)
+
+func Restart(extraArgs ...string) error {
+	execPath, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("failed to locate the executable file: %w", err)
+	}
+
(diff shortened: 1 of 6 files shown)


src/utils/ota/restart_windows.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	if len(args) > 0 {
+		cmdLine += " " + strings.Join(args, " ")
+	}
+
+	cmd := exec.Command("cmd.exe")
+	cmd.SysProcAttr = &syscall.SysProcAttr{CmdLine: cmdLine}
+
+	if err := cmd.Run(); err != nil {
```
Arriven, line 38: Doesn't `cmd.Run()` wait for completion btw? I feel like we're only interested in `cmd.Start()` here

Arriven (commented):
Looks good (for windows code xd), just have one minor question

sivillakonski:
@Arriven, what a sharp eye! I bet it was a last second change
Screenshot 2022-03-22 at 22 26 11

Arriven (approved):

PR #397, opened 2022-03-22
Update command line description

bitshape:
# Description

I think what you want to say is "primitive jobs are less resource intensive" or "more resource-efficient". The latter sounds wrong, so I'm going with the first one.

## Type of change

- [x] Documentation update

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -88,3 +88,7 @@
         but they cannot close it to their citizens
 
 ---
+
+???+ faq "What are primitive jobs?"
+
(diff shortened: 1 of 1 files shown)


Arriven:
Something like "primitive jobs rely on generating as much raw traffic as possible and may exhaust your system. They are also easier to detect and unadvisable to be used in clouds"

bitshape:
@Arriven 97c4f3dcbc17d3d9804661a69866d436bad41837

Arriven:
Markdown linter detected trailing spaces in the file

Arriven (approved):

PR #398, opened 2022-03-22
Added few cases for serialization_test

alexdanger:
# Description

Added test for `serialiation_test.go`, included check payload for ipv4/ipv6 packets and different payload string and int.

test cases

@Arriven just a question, is this ok to use some package for test like assert to avoid linter issues with cyclomatic complexity. I believe soon there is will be many type of packages and payload. So test cases will be a bit tricky to maintain.

### DIFF
--- a/src/core/packetgen/serialization_test.go
+++ b/src/core/packetgen/serialization_test.go
@@ -1,6 +1,8 @@
 package packetgen
 
 import (
+	"encoding/binary"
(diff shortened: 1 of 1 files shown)


Arriven (approved):

Arriven:
@alexdanger I don't have objections against testing frameworks - feel free to include them as long as they are maintained

PR #405, opened 2022-03-24
Make funlen even happier

jdoe7865623:


src/jobs/base.go
```diff
@@ -25,36 +25,51 @@ package jobs
...
+
 	ProxyURLs           string
-	ScaleFactor         int
 	SkipEncrypted       bool
-	Debug               bool
-	ClientID            string
 	EnablePrimitiveJobs bool
+	ScaleFactor         int
+
+	Debug bool
```
Arriven, line 49: I feel like this one is no longer used in jobs since I've switched debug logs to zap but it can be addressed in a separate PR

src/jobs/packetgen.go
```diff
@@ -76,41 +69,32 @@ func packetgenJob(ctx context.Context, logger *zap.Logger, _ GlobalConfig, args
...
-			logger.Debug("error sending packet", zap.Error(err))
-
-			return nil, err
+		n, err := rawConn.WriteTo(payloadBuf.Bytes(), nil, &net.IPAddr{IP: packet.IP()})
+		if err != nil {
+			return nil, fmt.Errorf("error sending packet: %w", err)
 		}
 
-		trafficMonitor.Add(uint64(len(payloadBuf.Bytes())))
+		trafficMonitor.Add(uint64(n))
```
Arriven, line 97: Ah, right, why didn't I do that myself? 😅

Arriven (approved):

PR #428, opened 2022-03-25
Get rid of advanced Docker image and change cloud deployments to use ENABLE_PRIMITIVE=false

### LINKED ISSUE #421: Change cloud guides and deployments to use regular image with ENABLE_PRIMITIVE=false
I want to get rid of advanced image bc it takes more than 20 minutes to build it (compared to 3-4 minutes when building regular one with ko). It's possible to make advanced image use ko for build as well but considering all the differences between advanced and regular images (just 1 param) having different images is an overkill

Amet13:
# Description

- Fix kubeval bug with linting all yamls in repo
- Cleanup all mentions for advanced image from repo
- Optimize k8s and terraform deployments to use `ENABLE_PRIMITIVE=false`

Fixes #421

## Type of change

- [x] Documentation update

## How Has This Been Tested?

```
$ helm upgrade --install \
    --create-namespace \
    --namespace=db1000n \
    -f values.yaml db1000n .
Release "db1000n" has been upgraded. Happy Helming!
NAME: db1000n
LAST DEPLOYED: Fri Mar 25 15:28:29 2022
NAMESPACE: db1000n
STATUS: deployed
REVISION: 3
TEST SUITE: None

$ kg po
NAME                       READY   STATUS    RESTARTS   AGE
db1000n-77b8f49569-j92q8   1/1     Running   0          24s

$ k apply -f daemonset.yaml 
namespace/db1000n unchanged
daemonset.apps/db1000n created
$ k apply -f deployment.yaml 
namespace/db1000n unchanged
deployment.apps/db1000n created
```


ansible/linux/setup.yaml
```diff
@@ -60,12 +60,12 @@
       shell: |
         echo "docker exec vpn expressvpn disconnect" >> /run.sh
         echo 'docker exec vpn expressvpn connect "$(shuf -n 1 /countries.txt)"' >> /run.sh
-        echo "docker run --name=db1000n --pull=always --net=container:vpn -e PUID=1000 -e PGID=1000 --rm -d ghcr.io/arriven/db1000n-advanced:latest" >> /run.sh
```
Amet13, line 63: Didn't set ENABLE_PRIMITIVE=false for Ansible only.
Should we add it as well?

Amet13: few mins

Dockerfile
```diff
@@ -11,15 +11,6 @@ ARG CA_PATH_VALUE
 ARG PROMETHEUS_BASIC_AUTH
 RUN make build_encrypted
 
-FROM alpine:3.15.2 as advanced
```
Arriven, line 14: I think you need to add this back too if you bring the action back :)

Amet13: definitely yes

PR #437, opened 2022-03-26
Reset metrics on config change

Kashkovsky:
# Description

When the config is updated with new targets, metrics remain the same hence meaningless.
My little proposal is to reset metrics on each config update to get relevant information about current targets.
Sorry if anything is wrong, I learned Go yesterday.

Fixes # (issue)
N/A
## Type of change

## How Has This Been Tested?

- Run the program with config A
- Update config
- Observe `Traffic` and `ProcessedTraffic` metrics are reset.

IMG_0070


Arriven:
Good idea, but can we make it a separate set of metrics instead? Ideally we'd show users both the general amount of traffic and traffic for latest targets as simple reset would probably be a bit confusing

Arriven:
Thinking about it more, I'll probably merge it as is since everything else works as if the app was restarted when config changes

src/runner/runner.go
```diff
@@ -240,3 +242,8 @@ func dumpMetrics(clientID string) error {
 
 	return nil
 }
+
+func resetMetrics(clientId string) {
```
Arriven, line 246:
```suggestion
func resetMetrics(clientID string) {
```
As per linter suggestion

src/runner/runner.go
```diff
@@ -240,3 +242,8 @@ func dumpMetrics(clientID string) error {
...
+	metrics.Default.Write(metrics.Traffic, clientId, 0)
```
Arriven, line 247: I can't remember whether this should be a client id or job id for it to work properly, did you test this? (I'll take a look in an hour or so myself, not near my PC yet)

Arriven (commented):
Some minor comments

src/runner/runner.go
Arriven: Yeah, I think you need to reset the tracker as a whole (lines 53-54). I think it's best to define that method on a storage type, trying to think whether it should be protected by some lock or not

Kashkovsky: Yep, I tested it with clientID, will move it to Storage, thank you.

Kashkovsky: I made the reset logic a method of Storage and reset them as a whole. Since metricsTracker uses concurrent sync.Map, I assume there’s no need to handle synchronization manually. Am I right?
Thinking further as a separate MR: that’s a good idea to make a separate metrics for targets. Maybe it will make sense to give a list of targets with their metrics bellow the aggregated values? So when a config is reset it’ll be pretty obvious for the user to recognise the meaning of the updated data.

Arriven: I'm almost sure that it should be safe to just replace a pointer but lets see if any linters object to that
As for metrics per target - be aware that we have support for encrypted targets and they shouldn't be shown to the user for security reasons

Kashkovsky: The only thing linter was complaining: line length in prometheus.go (fixed).

I see, considering encrypted targets, it will raise questions if we screen them in the output. If we omit them - aggregated output will become inconsistent with discrete metrics. Leaving it as is for now seems like the best choice

Arriven (approved):

PR #205, opened 2022-03-07
Added mirrors for config.

Devaniti:
# Description

Added mirrors for config files, in case Russia blocks github, so RU VPN users can still load config

Related to #204
Not closing since we may also add non git mirrors

## Type of change

Added additional links to mirrors of config files

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- When some links are purposufully broken, app correctly falls back to next ones

## Test Configuration

* Release version: 07e552eb9a740c44f88b44704df24e97c8ad6606
* Platform: Windows

## Logs

When all links are correct:
```text
C:\Projects\db1000n>go run main.go
2022/03/07 10:11:10.685127 runner.go:164: Loading config from "https://raw.githubusercontent.com/db1000n-coordinators/LoadTestConfig/main/config.json"
2022/03/07 10:11:10.685127 runner.go:87: New config received, applying
```
When some links are broken:
```text
C:\Projects\db1000n>go run main.go
2022/03/07 10:12:10.914883 runner.go:160: Failed to fetch config from "https://raw.qqqgithubusercontent.com/db1000n-coordinators/LoadTestConfig/main/config.json": Get "https://raw.qqqgithubusercontent.com/db1000n-coordinators/LoadTestConfig/main/config.json": dial tcp: lookup raw.qqqgithubusercontent.com: no such host
2022/03/07 10:12:10.915998 runner.go:160: Failed to fetch config from "https://qqqgitlab.com/bestdarkstar/LoadTestConfig/-/raw/main/config.json": Get "https://qqqgitlab.com/bestdarkstar/LoadTestConfig/-/raw/main/config.json": dial tcp: lookup qqqgitlab.com: no such host
2022/03/07 10:12:11.569429 runner.go:164: Loading config from "https://bitbucket.org/bestdark/loadtestconfig/raw/main/config.json"
2022/03/07 10:12:11.569429 runner.go:87: New config received, applying
```

PR #445, opened 2022-03-27
Make docker-compose not configurable

PerchunPak:
# Description

This will make docker-compose.yml not configurable. Only OpenVPN can be configured. Reason of this: I really hate resolve "merge-conflicts" when updating my docker-compose system. I just want use `git pull && docker-compose pull` to update without anything else. This PR make it possible, now you don't need disable half of file.

Cons:

- Breaking change.
- If your IP blocked in GitHub, you can't download config with `updater`. May be fixed with #205 (you really need fix?).

## Migrations

You need rename your `provider01.txt` (and/or `provider02.txt`) to `auth.txt`, also you need re-check your `provider*.conf` files, and configure it to work with one user (from `auth.txt`, insteand of `provider01.txt` or `provider02.txt`).
And you need rename `provider*.conf` to `provider*.ovpn` (example `provider01.endpoint02.conf` to `provider01.ovpn`).

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Documentation update

## How Has This Been Tested?

- [x] `docker-compose up -d`

## Test Configuration

- Release version: 779eb7fa43e72042ba35862e29e496b8903901d3 with `v0.8.10`
- Platform: Any, which support docker.

## Logs

- [x] No logs.

## Screenshots

- [x] No screenshot


Arriven:
Ideally I'd want to have both this static version and old example of having multiple different providers present. Can you move that old version to the examples folder? (Also while you're at it could you change config/examples to examples/config so that everything would use that same pattern?)

PerchunPak:
Moving this in example folder will ruin all sense of this PR. I want make updates with just few commands, instead of moving all edited files and after pull, move it back. Also there is only one feature that this config not support comparing with previous - `provider02*` files, but you really need few programs runned at once? It is actually have no sence, because one will be more effective and optimised.
Oh, and one more feature that I can't realise in static configuration - optional `updater`. But as I mention above:
Untitled

Arriven:
This docker compose doesn't depend on any files in the repo except those provider ones so you can have that immutable use case regardless of the folder it's in (yeah, you'd need to have one initial migration if we do move the folder but the risk of that happening again is low). The reason I want to move this to a folder is because the repo grows and becomes harder to manage if everything is at root level and grouping related things together helps with that (I also want to move terraform and kubernetes there following the same logic)
As for keeping the old version of this compose file there is a benefit of having multiple vpn providers - if one goes temporarily down you still have others + it allows to disperse the traffic and make it harder to block attacks based on IP/geo so I want to at least have some reference for people on how to achieve that

PerchunPak:
> benefit of having multiple vpn providers - if one goes temporarily down you still have others + it allows to disperse the traffic and make it harder to block attacks based on IP/geo

This will work normaly only if tool will have build-in feature for proxies support, but seems like someone deleted it. Also 2 VPN providers not achieve this milestone, and increase this integer will just increese CPU consumption in geometry progression.

Anyway I will move it to `examples` folder.

Arriven (approved):
Not exactly what I meant (I thought you'd move ovpn folder so that it stays the same relative to the compose file) but this also works

Arriven:
As for 1 vpn vs multiple vpns vs proxies - proxylist is the easiest one to setup globally but it only supports some of attacks (I'm not sure if you can send upd or raw ip traffic through it, at least using standard go library). VPN doesn't have that issue and multiple vpns are just a way for someone to make the vpn setup more robust. The best way is to actually deploy the app to multiple small cloud machines but that's a separate discussion

PR #484, opened 2022-04-03
Implement the new metrics Reporter

jdoe7865623:
This is a work in progress for feedback, migration to the new reporter is not done yet.

The new console stats will tentatively look as follows:

```
 --- Traffic stats ---
 |            Target | Requests attempted | Requests sent | Responses received | Data sent | Data received |
 | http://foobar.com |               1100 |           110 |                 11 |  17.19 MB |       4.30 MB |
 |  tcp://foobar.com |               1100 |           110 |                 11 |   1.72 MB |       0.43 MB |
 |               --- |                --- |           --- |                --- |       --- |           --- |
 |             Total |               2200 |           220 |                 22 |  18.91 MB |       4.73 MB |
```


roman-kruglov:
Hi! I thought about something similar too, to better inform the user about the results and thus efficiency. So I had one more thought - maybe it should somehow specify summarised response codes and something like closed / rejected connections. And maybe some efficiency coefficient. No idea how to calculate it, but it would be great - e.g. to try different VPNs / servers. What do you think? (yes, I'm willing to participate in coding, just trying to cooperate because I found your pr which seems close to something I would do, so I decided to write here first).

roman-kruglov:
Also had an idea to replace those tons of messages about starting a thread for a target like
```
2022/04/06 17:07:48.414177 http.go:147: Attacking https://kupiprodai.ru
2022/04/06 17:07:48.414892 http.go:147: Attacking https://kupiprodai.ru
2022/04/06 17:07:48.415018 http.go:147: Attacking http://kupiprodai.ru
2022/04/06 17:07:48.415130 http.go:147: Attacking https://kupiprodai.ru
2022/04/06 17:07:48.415163 http.go:147: Attacking http://kupiprodai.ru
```
with something more combined, like maybe
```
2022/04/06 17:07:48.414177 http.go:147: Attacking https://kupiprodai.ru (30 threads)
2022/04/06 17:07:48.414892 http.go:147: Attacking https://rudos.ru (20 threads)
etc
```
would be much shorter and easier to follow

jdoe7865623:
> Hi! I thought about something similar too, to better inform the user about the results and thus efficiency. So I had one more thought - maybe it should somehow specify summarised response codes and something like closed / rejected connections. And maybe some efficiency coefficient. No idea how to calculate it, but it would be great - e.g. to try different VPNs / servers. What do you think? (yes, I'm willing to participate in coding, just trying to cooperate because I found your pr which seems close to something I would do, so I decided to write here first).

Fair enough; we can't summarize by response codes since there are different protocols involved - not only HTTP. This is why I thought of the five metrics you can see in the table as a common denominator of sorts. Collecting these metrics from all supported jobs is still a WIP.

You may define "efficiency" as a ratio between the responses received and requests attempted (or sent). But this is somewhat subjective (what if the target is down due to the swarm activity? :), so I decided to start with plain stats. I'm open to suggestions, though.

jdoe7865623:
> Also had an idea to replace those tons of duplicated repetitive messages about starting a thread for a target like
> 
> ```
> 2022/04/06 17:07:48.414177 http.go:147: Attacking https://kupiprodai.ru
> 2022/04/06 17:07:48.414892 http.go:147: Attacking https://kupiprodai.ru
> 2022/04/06 17:07:48.415018 http.go:147: Attacking http://kupiprodai.ru
> 2022/04/06 17:07:48.415130 http.go:147: Attacking https://kupiprodai.ru
> 2022/04/06 17:07:48.415163 http.go:147: Attacking http://kupiprodai.ru
> ```
> 
> with something more combined and shorter, like maybe
> 
> ```
> 2022/04/06 17:07:48.414177 http.go:147: Attacking (30 threads) https://kupiprodai.ru
> 2022/04/06 17:07:48.414892 http.go:147: Attacking (20 threads) https://rudos.ru
> etc
> ```
> 
> would be much shorter and easier to follow

Yeah, those are not super-helpful at this point. The runner does not know much about the specific job config, though, thus we log this stuff from within the jobs in coroutines. That causes a barrage of identical messages. Feel free to submit a PR if you come up with a smart suggestion :)

roman-kruglov:
ok, yep, I'll try. I have several small ideas to start from.

src/job/http.go
```diff
@@ -71,9 +71,14 @@ func singleRequestJob(ctx context.Context, args config.Args, globalConfig *Globa
 		log.Printf("Sent single http request to %v", requestConfig.Path)
 	}
 
-	// dataSize :=
 	http.InitRequest(requestConfig, req)
 
+	// dataSize, _ :=
+	_, err = req.WriteTo(metrics.NopWriter{})
```
Arriven: you can just remove this if you don't need to calculate data size

jdoe7865623: That wasn't the final version. You can take a look at the http job now.

src/utils/metrics/metrics.go
```diff
@@ -20,113 +20,289 @@
...
 }
 
-// Flush used to flush pending metrics updates to the storage
-func (w *Writer) Flush() {
-	w.ms.Write(w.name, w.jobID, w.value)
+// Inc increases Accumulator Stat value by 1. Returns self for chaining.
+func (a *Accumulator) Inc(target string, s Stat) *Accumulator { return a.Add(target, s, 1) }
+
+// AddStats to the Accumulator. Returns self for chaining.
+func (a *Accumulator) AddStats(target string, s Stats) *Accumulator {
```
Arriven, line 258: Can we have a function that would expect a variadic list of pairs <Stat, value> instead? The way this API is implemented now it's easy to misuse it by passing a slice of wrong size or reordering arguments accidentally

src/utils/metrics/metrics-old.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+}
+*/
+
+// NopWriter implements io.Writer interface to simply track how much data has to be serialized
+// type NopWriter struct{}
+
+// func (w NopWriter) Write(p []byte) (n int, _ error) {
+// 	return len(p), nil
+// }
```
Arriven, line 134: can you delete this file completely if you are commenting it out anyway?

src/utils/metrics/metrics.go
```diff
@@ -20,113 +20,289 @@
...
+
+	clientID string
+}
+
+// NewReporter creates a new Reporter.
+func NewReporter(clientID string) *Reporter { return &Reporter{clientID: clientID} }
+
+// Sum returns a total sum of metric s.
+func (r *Reporter) Sum(s Stat) uint64 {
+	if r == nil || !s.valid() {
```
Arriven, line 80: What is the use case for these checks? We don't have any user-defined data coming into this API so it would be better to panic here to catch issues faster and have a clear branch prediction pattern for the compiler that should allow for better optimizations

Arriven: I initially assumed you want this check for a case of disabling metrics for encrypted jobs but I see that you still pass a correct accumulator handle there so it shouldn't be needed

Arriven: same obviously goes for all other checks of this type

src/utils/metrics/metrics-old.go
jdoe7865623: ah, right

src/utils/metrics/metrics.go
jdoe7865623: Yes, the intention was to disable collection for encrypted jobs. I must have forgotten about this detail, let me cross-check.

As for panicking I also tend to agree, guess I'll just drop the validations.

src/utils/metrics/metrics.go
jdoe7865623: See the `NewStats` constructor instead which should solve your concern

Arriven (approved):

PR #487, opened 2022-04-04
Pass an URL to a proxy list using `proxy-path` flag

### LINKED ISSUE #475: Proxy not work for Windows
## Expected Behavior
work

## Actual Behavior
win_error

## Steps to Reproduce the Problem

  .\db1000n.exe --proxy '{{ join (split (get_url "https://raw.githubusercontent.com/porthole-ascend-cinnamon/proxy_scraper/main/proxies.txt") "\n") "," }}'

## Specifications

  - Version: 0.8.15
  - Platform: Windows 10 x64
https://github.com/Arriven/db1000n/issues/424

Kashkovsky:
Fixes #475

## Type of change
- [x] Documentation update

## How Has This Been Tested?

```
./db1000n -proxy-path https://raw.githubusercontent.com/porthole-ascend-cinnamon/proxy_scraper/main/proxies.txt
```

### DIFF
--- a/docs/advanced-docs/configuration.md
+++ b/docs/advanced-docs/configuration.md
@@ -26,7 +26,9 @@ Usage of db1000n:
   -prometheus_on
       Start metrics exporting via HTTP and pushing to gateways (specified via <prometheus_gateways>) (default true)
   -proxy string
-      system proxy to set by default (can be a comma-separated list or a template)
+      system proxy (SYSTEM_PROXY environment variable) to set by default (can be a comma-separated list or a template)
+  -proxy-path string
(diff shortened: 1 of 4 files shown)


Arriven:
> This particular approach is almost the same as downloading that proxylist, reformatting it with simple commandline and passing it directly into the command. Current approach re-downloads the proxylist each time the job is restarted (right now it's only when the config is updated but still)

What would be better is to provide template functions like `split_lines` and `join_list` (naming can be different, it's just first names that came to my head after 30 seconds of thinking) that would remove the need to escape those sequences

PR #517, opened 2022-04-24
Add a new flag `-log-format` and a `simple` format

roman-kruglov:
the most human readable format for those who only look at the output in a terminal

# Description

With the recent switch to `zap` as the used logger it became much harder to just look at terminal's output. If you just run it manually and the output is for reading rather than logging, using `-log-format simple` makes the output much more bearable.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

By running and looking at the output in a terminal.

## Test Configuration

- Release version: latest main branch (v0.8.30 is the latest current release)
- Platform: Mac OS

## Logs

 Output sample (real values crossed out)
```text
db1000n % ./db1000n -log-format simple
running db1000n {"version": "v0.8.30", "pid": 84370}
checking IP address,    {"iter": 1}
error fetching location info    {"error": "Get \"https://sample/\": context deadline exceeded"}
checking IP address,    {"iter": 2}
location info   {"country": "Some country", "ip": "XX.XX.XX.XX"}
loading config  {"path": "sample"}
new config received, applying
attacking       {"target": "http://sample"}
attacking       {"target": "http://sample"}
attacking       {"target": "http://sample"}
error running job       {"name": "", "type": "encrypted", "error": "encryption not supported"}
job instances (re)started       {"count": 874}
attacking       {"target": "https://sample/sample"}
single http request     {"target": "https://sample/"}
```

PR #518, opened 2022-04-24
Combine metrics summary into one object

roman-kruglov:
so the whole report goes as one log entry with a JSON object for each target

# Description

This might be helpful for logging and assessing to have the whole summary report as one entry. Though arguably it makes it even less readable in terminal's output, and I tried to improve that with pr #517. I tried to make `zap` pretty print this huge JSON report object (for the case when `-log-format simple` from #517), but it was way too complicated for such a simple feature. Though it might still be possible.

What do you guys think? Could this be helpful?

## Type of change

Potentially somebody could rely on the current summary format, but with so many swift changes and the recent move to `zap` - I doubt it.

## How Has This Been Tested?

By running and looking at the output in a terminal.

## Test Configuration

- Release version: latest main branch (v0.8.30 is the latest current release)
- Platform: Mac OS

## Logs

Output sample (real values crossed out, truncated for shorter text)

```text
{"http://sample": {"requests_attempted": 7, "requests_sent": 0, "responses_received": 0, "bytes_sent": 0}, "http://XX.XX.XX.XX": {"requests_attempted": 7, "requests_sent": 0, "responses_received": 0, "bytes_sent": 0}, "http://sample": {"requests_attempted": 7, "requests_sent": 0, "responses_received": 0, "bytes_sent": 0}, "total": {"requests_attempted": 185664, "requests_sent": 183986, "responses_received": 138, "bytes_sent": 1862621}}
```
the full log entry with the current default format would look like this:
```text
{"level":"info","ts":1650809773.640194,"caller":"metrics/metrics.go:183","msg":"stats","http://sample":{"requests_attempted":3,"requests_sent":3,"responses_received":3,"bytes_sent":624},"http://sample":{"requests_attempted":1,"requests_sent":1,"responses_received":1,"bytes_sent":191},"http://sample":{"requests_attempted":1,"requests_sent":1,"responses_received":1,"bytes_sent":197},"tcp://XX.XXX.X.XX:XX":{"requests_attempted":91924,"requests_sent":91924,"responses_received":0,"bytes_sent":919240},"total":{"requests_attempted":184496,"requests_sent":183902,"responses_received":54,"bytes_sent":1848272}}
```

### DIFF
--- a/src/utils/metrics/metrics.go
+++ b/src/utils/metrics/metrics.go
@@ -28,6 +28,7 @@ import (
 	"sync"
 
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 // Stat is the type of statistical metrics.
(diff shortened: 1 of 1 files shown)


Arriven (approved):

roman-kruglov:
I have several more ideas regarding this and simple output too, I'll have more improvements

Arriven:
I've added a change that groups up all the targets into an object to simplify parsing (you can safely iterate on all keys in targets json object instead of having to filter root keys), please be aware

roman-kruglov:
> I've added a change that groups up all the targets into an object to simplify parsing (you can safely iterate on all keys in targets json object instead of having to filter root keys), please be aware

yep, was thinking about combining targets for this same purpose too

PR #519, opened 2022-04-26
Simple output improvements (also make simple output format default)

roman-kruglov:
# Description

I want to improve it further to show a pretty summary like before. For this I split `Reporter` into several implementations - for zap and console.

 Also `simple` output format is made default now.

It's a work in progress - I intend to split `mectrics.go` into several files. But I hope it gives a preview of what's coming if somebody touches same files.

**P.S.** I intend to finish this somewhere around today's evening.

## Type of change

if some scripts or bots use json as default now - we need to update those before releasing - the default becomes simple.

## How Has This Been Tested?

By running and looking at the output in a terminal.

## Test Configuration

- Release version: latest main branch (v0.8.31 is the latest current release)
- Platform: Mac OS

## Logs

Output sample (real values crossed out, truncated for shorter text, alignment in console is better)

```text
single http request     {"target": "https://sample/"}
single http request     {"target": "https://sample"}

 --- Traffic stats ---
 |                                Target | Requests attempted | Requests sent | Responses received | Data sent |
 |                  http://xx.xx.xx.xx |                  1 |                   0 |                  0 |     0.00 MB |
 |                  http://xx.xx.xx.xx |                  3 |                  0 |                  0 |     0.00 MB |
 |                        http://sample |                  2 |                  0 |                  0 |     0.00 MB |
 |                      https://sample |                  1 |                   0 |                  0 |     0.00 MB |
 |                                         --- |               --- |                --- |               --- |              --- |
 |                                      Total |   14623715 |    14620788 |           1490 | 139.66 MB |

loading config  {"path": "https://sample/config.json"}
the config has not changed. Keep calm and carry on!
```

### DIFF
--- a/src/job/runner.go
+++ b/src/job/runner.go
@@ -65,13 +65,15 @@ func NewConfigOptionsWithFlags() *ConfigOptions {
 type Runner struct {
 	cfgOptions    *ConfigOptions
 	globalJobsCfg *GlobalConfig
+	reporter      metrics.Reporter
 }
 
 // NewRunner according to the config
-func NewRunner(cfgOptions *ConfigOptions, globalJobsCfg *GlobalConfig) *Runner {
+func NewRunner(cfgOptions *ConfigOptions, globalJobsCfg *GlobalConfig, reporter metrics.Reporter) *Runner {
 	return &Runner{
 		cfgOptions:    cfgOptions,
 		globalJobsCfg: globalJobsCfg,
+		reporter:      reporter,
 	}
 }
 
@@ -84,10 +86,7 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger) {
 	defer refreshTimer.Stop()
 	metrics.IncClient()
 
-	var (
-		cancel   context.CancelFunc
-		reporter *metrics.Reporter
-	)
+	var cancel context.CancelFunc
 
 	for {
 		rawConfig := config.FetchRawMultiConfig(logger, strings.Split(r.cfgOptions.PathsCSV, ","),
@@ -105,14 +104,13 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger) {
 				cancel()
 			}
 
+			r.reporter.ClearData() // clear info about previous targets
+
 			if rawConfig.Encrypted {
 				logger.Info("config is encrypted, disabling logs")
-
-				reporter = nil
-				cancel = r.runJobs(ctx, cfg, reporter, zap.NewNop())
+				cancel = r.runJobs(ctx, cfg, nil, zap.NewNop())
 			} else {
-				reporter = metrics.NewReporter(r.globalJobsCfg.ClientID)
-				cancel = r.runJobs(ctx, cfg, reporter, logger)
+				cancel = r.runJobs(ctx, cfg, r.reporter, logger)
 			}
 		} else {
 			logger.Info("the config has not changed. Keep calm and carry on!")
@@ -129,8 +127,8 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger) {
 			return
 		}
 
-		if reporter != nil {
-			reportMetrics(reporter, r.globalJobsCfg.ClientID, logger)
+		if r.reporter != nil {
+			reportMetrics(r.reporter, r.globalJobsCfg.ClientID, logger)
 		}
 	}
 }
@@ -151,7 +149,7 @@ func nonNilConfigOrDefault(c, defaultConfig *config.RawMultiConfig) *config.RawM
 	return defaultConfig
 }
 
-func (r *Runner) runJobs(ctx context.Context, cfg *config.MultiConfig, reporter *metrics.Reporter, logger *zap.Logger) (cancel context.CancelFunc) {
+func (r *Runner) runJobs(ctx context.Context, cfg *config.MultiConfig, reporter metrics.Reporter, logger *zap.Logger) (cancel context.CancelFunc) {
 	ctx, cancel = context.WithCancel(ctx)
 
 	var jobInstancesCount int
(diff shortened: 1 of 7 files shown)


src/utils/metrics/reporter.go
```diff
@@ -0,0 +1,120 @@
...
+	ClearData()
+
+	// NewAccumulator returns a new metrics Accumulator for the Reporter.
+	NewAccumulator(jobID string) *Accumulator
+}
+
+// ZapReporter
+
+type ZapReporter struct {
+	data   ReporterData
```
Arriven, line 27: You can use type embedding to avoid specifying methods that just call the same method of data

src/utils/metrics/reporter.go
```diff
@@ -0,0 +1,120 @@
...
+	"go.uber.org/zap"
+)
+
+// Reporter gathers metrics across jobs and reports them.
+// Concurrency-safe.
+type Reporter interface {
+	// WriteSummary dumps Reporter contents into the target.
+	WriteSummary()
+
+	SumBytesSent() uint64
```
Arriven, line 17: I think it's better to have generic Sum(stat) interface in case we have usage for other metrics outside the package

src/job/runner.go
```diff
@@ -105,14 +104,14 @@ func (r *Runner) Run(ctx context.Context, logger *zap.Logger) {
 				cancel()
 			}
 
+			r.reporter.ClearData() // clear info about previous targets
```
Arriven, line 107: The main reason reporter was recreated with each new config is so that jobs that haven't been killed right away wouldn't have a chance to report their metrics to the new iteration. This approach may bring that bug back. Can we instead split the reporter interface into actual reporting/logging part and storage part? Then reporter could have a single method for writing summary that accepts a storage and we would be able to recreate storage on each new config

src/utils/metrics/reporter.go
```diff
@@ -0,0 +1,120 @@
...
+	target io.Writer
+}
+
+// NewConsoleReporter creates a new Reporter which outputs straight to the console
+func NewConsoleReporter(target io.Writer) Reporter {
+	return &ConsoleReporter{target: target}
+}
+
+func (r *ConsoleReporter) WriteSummary() {
+	w := tabwriter.NewWriter(r.target, 1, 1, 1, ' ', tabwriter.AlignRight)
```
Arriven, line 71: To be honest, I'm not exactly fond of this tab writer interface, especially since all the other output is done via single library. If the format doesn't suffer too much I'd rather have this reporter print stats via zap similar to how it was done before your previous PR that grouped the stats together

Arriven: Main motivation here is to simplify maintenance

Arriven (commented):
Looks good, but there are couple things that need to be addressed first

src/job/runner.go
roman-kruglov: Actually, it's already split - `Reporter` interface with two implementations and `ReporterData` (which I'll probably rename to `Metrics` or something) struct with all info. Yep, it would be even easier not to store it inside I guess.. I'll look into it

src/utils/metrics/reporter.go
roman-kruglov: my motivation was - to make terminal output (if nobody actually logs) readable again - it's really hard with `zap` - they don't even support indentation for json. There are ways.. but it's like reimplementing a whole `Encoder`. Or wrapping. In short - I didn't find an easy way to pretty-print into something readable with `zap`. There are also external tools, but those are separate binaries and they won't be able to eat this simplified output anyway.

So the main problem for me personally is now - I can't tell anymore if load testing is going well or if all requests are blocked. That influenced my choice of VPNs, etc. That's why I was trying to resurrect the pretty table - it was soooo helpful with this. It would be just one method after all.

So, do you think I should go the hard way and attempt it with zap? It will probably be much more code.

AndyKrivovjas: @Arriven 
I miss the table output as well. I run db1000n in docker containers, each one represents a separate VPN connection. It was relatively easy to print the logs and see the actual response rate for each instance. Having an option to print the table would be great (maybe introduce a new flag). Or maybe create a service, which can parse the output and display the stats (could be a docker container running a web server and a simple dashboard).

roman-kruglov: yep, in my case no zap output with stats is ever readable. It was a bit more readable when every target was logged as a separate log entry, but it made more sense to combine the whole stats output into one entry for analysis, which I did.

"we output each target in a different log entry with each field being separate" - I think it was Before the change when I combined all stats into one entry, and you then brushed it up a bit and now it is something like this:
`logger.Info("stats", zap.Object("total", &totals), zap.Object("targets", stats))`

But even before combining all stats into one entry it was only somewhat readable for me - even with a separate log entry for each target. Nothing like the sweet table.

src/utils/metrics/reporter.go
roman-kruglov: after complete separation of metrics data from reporters I didn't even have to, and the code became shorter. Well, at least I learned what those type embeddings are

src/utils/metrics/reporter.go
roman-kruglov: Exposed the old full Sum() function

src/job/runner.go
roman-kruglov: Split it, now Metrics are about holding combined stats and Reporter is about reporting. The bug should be prevented now, I create a new storage on a new config.

src/utils/metrics/reporter.go
roman-kruglov: I kept the table for now, after the separation the whole table reporter implementation is only ~+20 lines of code working with a tab writer, and it uses the separated metrics data storage for all data acquisition. I hope it's not such a big overhead for the feature many loved and used.. but if you insist - I'll remove it.

roman-kruglov:
Guys, could somebody look through the pr with a fresh eye to make sure I didn't fuck up memory somewhere. Pretty exhausted right now, I'll skim over it again in the morning.

Arriven (approved):

PR #530, opened 2022-05-04
Fix double total stats reported in table simple output

roman-kruglov:
# Description

When run with simple output format the table reported doubled total stats. There was an additional loop for totals which I didn't notice when copied table's formatter implementation. Also I shrank the table implementation a bit.

## Type of change

## How Has This Been Tested?

By running in a terminal

## Test Configuration

- Release version: latest dev (current release version is .33)
- Platform: Mac OS

### DIFF
--- a/src/utils/metrics/reporter.go
+++ b/src/utils/metrics/reporter.go
@@ -44,37 +44,33 @@ func NewConsoleReporter(target io.Writer) Reporter {
 }
 
 func (r *ConsoleReporter) WriteSummary(metrics *Metrics) {
-	w := tabwriter.NewWriter(r.target, 1, 1, 1, ' ', tabwriter.AlignRight)
-
-	defer w.Flush()
-
+	writer := tabwriter.NewWriter(r.target, 1, 1, 1, ' ', tabwriter.AlignRight)
 	stats, totals := metrics.SumAllStats()
 
-	fmt.Fprintln(w, "\n --- Traffic stats ---")
-	fmt.Fprintf(w, "|\tTarget\t|\tRequests attempted\t|\tRequests sent\t|\tResponses received\t|\tData sent \t|\n")
+	defer writer.Flush()
 
-	const BytesInMegabyte = 1024 * 1024
+	// Print table's header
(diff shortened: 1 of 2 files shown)


src/utils/metrics/reporter.go
```diff
@@ -44,37 +44,33 @@ func NewConsoleReporter(target io.Writer) Reporter {
...
-		fmt.Fprintf(w, "|\t%s\t|\t%d\t|\t%d\t|\t%d\t|\t%.2f MB \t|\n", tgt,
-			tgtStats[RequestsAttemptedStat],
-			tgtStats[RequestsSentStat],
-			tgtStats[ResponsesReceivedStat],
-			float64(tgtStats[BytesSentStat])/BytesInMegabyte,
-		)
-
-		for s := range totals {
-			totals[s] += tgtStats[s]
-		}
```
roman-kruglov, line 70: this was the actual loop which doubled the total stats - now all calculations are done by `stats, totals := metrics.SumAllStats()` call and there is no need for reporters to calculate anything

src/utils/metrics/reporter.go
```diff
@@ -44,37 +44,33 @@ func NewConsoleReporter(target io.Writer) Reporter {
...
-	w := tabwriter.NewWriter(r.target, 1, 1, 1, ' ', tabwriter.AlignRight)
-
-	defer w.Flush()
-
+	writer := tabwriter.NewWriter(r.target, 1, 1, 1, ' ', tabwriter.AlignRight)
 	stats, totals := metrics.SumAllStats()
 
-	fmt.Fprintln(w, "\n --- Traffic stats ---")
-	fmt.Fprintf(w, "|\tTarget\t|\tRequests attempted\t|\tRequests sent\t|\tResponses received\t|\tData sent \t|\n")
+	defer writer.Flush()
```
Arriven, line 50: why did you move flush after SumAllStats()?

roman-kruglov: just to combine variable declarations together

Arriven (approved):

PR #531, opened 2022-05-04
Fix a crash when interface name is empty

roman-kruglov:
# Description

Main's top crashes with lots of errors about dereferencing nil. Happens in Bind(). Because `getSockaddrByName()` can return `nil` in lots of cases. Not sure if I fixed it entirely correctly, but this is my take on it. Maybe instead of returning nil this route needs to evade something above when interface's name is empty.

## Type of change

## How Has This Been Tested?

Ran in my terminal

## Test Configuration

- Release version: main's top (latest release is .33)
- Platform: Mac OS

### DIFF
--- a/src/utils/utils_unix.go
+++ b/src/utils/utils_unix.go
@@ -60,11 +60,16 @@ func getSockaddrByName(name string) syscall.Sockaddr {
 }
 
 func BindToInterface(name string) func(network, address string, conn syscall.RawConn) error {
+	sockAddr := getSockaddrByName(name)
+	if sockAddr == nil {
+		return nil
+	}
+
 	return func(network, address string, conn syscall.RawConn) error {
 		var operr error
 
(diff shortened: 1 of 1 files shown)


src/utils/utils_unix.go
```diff
@@ -60,11 +60,16 @@ func getSockaddrByName(name string) syscall.Sockaddr {
 }
 
 func BindToInterface(name string) func(network, address string, conn syscall.RawConn) error {
+	sockAddr := getSockaddrByName(name)
```
Arriven, line 63: ah, right, although this has to be done in the returned function to work in case network changes (i.e. when you're on wifi)

I'll address myself

Arriven: properly fixed in https://github.com/Arriven/db1000n/commit/9924a9d84168c5b3d112c3c3796b5c5ad6ad3dac

Issue #528, opened 2022-05-04
Change country checker to use fasthttp with all network settings (proxy, interface, etc.)


Arriven:
Works with ipv6 ! But still small bug that was not present at pre 0.8x series - countrycheker still uses ipv4 addressing for country check :

```
utils/countrychecker.go:78      location info   {"country": ....
```

_Originally posted by @Sfinx in https://github.com/Arriven/db1000n/issues/521#issuecomment-1116467348_

Arriven:
fixed in 060ae19ad8805ef0b7e0b161ac5a1a6f1df6e086

Issue #534, opened 2022-05-06
SkipBody - what it really does?


deputinizer:
Well, SkipBody does something different I think. 
```golang
resp.SkipBody = true
```
https://github.com/Arriven/db1000n/commit/3147170641eb0c3da75cfc74d68d9606d4e23d5a

The server sends body response but we are ignoring it, thus mishandling the HTTP protocol.
https://github.com/valyala/fasthttp/blob/7cc6f4c513f9e0d3686142e0a1a5aa2f76b3194a/http.go#L1305

I suppose what you really wanted is like:
```golang
resp.DiscardBody = true
```
https://github.com/erkexzcx/stoppropaganda/blob/main/internal/stoppropaganda/websites.go#L286

I've already implemented it in stoppropaganda by editing fasthttp.
https://github.com/deputinizer/fasthttp/commits/master

I'm open to discussion, maybe I'm wrong... but HTTP protocol is pretty simple I think. :wink:

Arriven:
the idea was to have a way to utilize cpu better by not needing to do unnecessary work if the goal is to just spam the target with traffic. I then realized that it's better done by utilizing packetgen but forgot to remove this part, thanks for reminding!

Issue #533, opened 2022-05-06
Why is db1000n so slow...


deputinizer:
![image](https://user-images.githubusercontent.com/100740281/167130319-13cc00bb-4cb0-4714-b4e0-1cad7bfb8008.png)

I'm pressing ctrl+C

deputinizer:
And why is this metrics thing so retarded
![image](https://user-images.githubusercontent.com/100740281/167130812-e97730eb-3761-4950-84e6-1a126671b9eb.png)

PR #535, opened 2022-05-06
Performance + protocol fixes

deputinizer:
# Description

Performance fixes + HTTP 1.1 protocol fix.

Fixes 
- #533 
- #534
- #528

## Type of change

- [x] Bug fix: ctrl+C slowly responding even without prometheus metrics
- [x] New feature: tls.Config cache in metrics
- [x] Breaking change / fix: SkipBody did unintended things (makes TCP reconnect because of malformed protocol)
- [x] Breaking change: added Custom resolver from stoppropaganda
- [x] Breaking change: countrychecker using fasthttp

## Tested

Tested with pprof CPU profile (`svg` command).
```console
ubuntu@pc:~$ ./db1000n -pprof :6060 -log-level debug 2> logs.txt
...
ubuntu@pc:~$ go tool pprof http://localhost:6060/debug/pprof/profile?seconds=60
```
## Logs
So i've added some more `dialing` debug (not in commits :smiley: )
Multiple HTTP requests on one connection proves that #534 is correct

```bash
grep -v "dial tcp" logs.txt > logs_no_dial.txt
```

With `resp.SkipBody = true`
```text
single fasthttp job request	{"atomic_id": 1452, "target": "https://kim.tektorg.ru/"}
dialing	{"addr": "kim.tektorg.ru:443"}
single fasthttp job request	{"atomic_id": 1453, "target": "https://kim.tektorg.ru/"}
dialing	{"addr": "kim.tektorg.ru:443"}
single fasthttp job request	{"atomic_id": 1454, "target": "https://fin.tektorg.ru/bg"}
dialing	{"addr": "fin.tektorg.ru:443"}
single fasthttp job request	{"atomic_id": 1455, "target": "https://fin.tektorg.ru/bg"}
dialing	{"addr": "fin.tektorg.ru:443"}
single fasthttp job request	{"atomic_id": 1456, "target": "https://fin.tektorg.ru/bg"}
dialing	{"addr": "fin.tektorg.ru:443"}
single fasthttp job request	{"atomic_id": 1457, "target": "https://fin.tektorg.ru/bg"}
dialing	{"addr": "fin.tektorg.ru:443"}
single fasthttp job request	{"atomic_id": 1458, "target": "https://fin.tektorg.ru/bg"}
dialing	{"addr": "fin.tektorg.ru:443"}
single fasthttp job request	{"atomic_id": 1459, "target": "https://fin.tektorg.ru/bg"}
dialing	{"addr": "fin.tektorg.ru:443"}
```

With `resp.SkipBody = false`
```text
single fasthttp job request	{"atomic_id": 1358, "target": "https://kim.tektorg.ru/"}
single fasthttp job request	{"atomic_id": 1359, "target": "https://kim.tektorg.ru/"}
single fasthttp job request	{"atomic_id": 1360, "target": "https://www.tektorg.ru/"}
single fasthttp job request	{"atomic_id": 1361, "target": "https://kim.tektorg.ru/"}
single fasthttp job request	{"atomic_id": 1362, "target": "https://irao.tektorg.ru/"}
single fasthttp job request	{"atomic_id": 1363, "target": "https://irao.tektorg.ru/"}
single fasthttp job request	{"atomic_id": 1364, "target": "https://www.tektorg.ru/"}
single fasthttp job request	{"atomic_id": 1365, "target": "https://www.tektorg.ru/"}
```

As you can see `SkipBody == true` breaks the protocol and makes TLS handshakes constantly after every single HTTP request.

## Screenshots

- No screenshots because `tls.Config` decryption disappeared in CPU profile :shrug: (that's good actually :+1: )


deputinizer:
Lol why use lint :sob: 
Lint disallows `this` and `self`

deputinizer:
We can make DNS resolver use almost 0% CPU
image

deputinizer:
Works.
image

However, I think this linter thing is bullshit

deputinizer:
Also, stoppropaganda + db1000n authors should do some collab and on rewriting the whole thing in Rust.

deputinizer:
These templates are not entirely necessary...
<details>
  <summary>Spoiler: pprof memory</summary>
  
image

</details>

deputinizer:
Probably fixed #528

src/core/customresolver/customresolver.go
```diff
@@ -0,0 +1,75 @@
+package customresolver
```
Arriven, line 1: I don't like this change for multiple reasons:
- it's a bit over engineered. I can see what it does with caching and sharing that cache between goroutines but it has too much code just for that
- it's hardcoded just for fight with russian propaganda. I actually mean it when I say that this project can be used by actual red teams to load test their own infra. This change would make it harder
- going back to performance benefits of caching - you'll actually get max performance benefits by not doing any dns lookups at all and it already allows you to do that by resolving your dns outside db1000n (see statil_host in http client). This will allow you even more than dns caching as it will work across processes on different machines and also there are types of defense that can be bypassed by bypassing dns if improperly configured. This is what we do in the default config and there's a lot of recon going to that part that is not visible here
- Finally if you do want to have a way to dynamically query dns while running the app but share the results between multiple goroutines it's better to be achieved by having a "lookup_host" kind of template (I'm not sure if I don't have it already) and using more complex scenario in the config (storing dns result in a variable and using it in a sequence). You can even make a version of the template that would accept the list of dns servers to use. This makes the least amount of assumptions about average target and leaves most control capabilities to the one configuring the app

src/core/spdnsclient/net.go
```diff
@@ -0,0 +1,57 @@
+package spdnsclient
```
Arriven, line 1: Same comment as in `customdial`

src/job/http.go
```diff
@@ -149,7 +151,8 @@ func fastHTTPJob(ctx context.Context, args config.Args, globalConfig *GlobalConf
 		fasthttp.ReleaseResponse(resp)
 	}()
 
-	resp.SkipBody = true
+	// TODO: DiscardBody here
+	resp.SkipBody = false
```
Arriven, line 155: If I'm not mistaken this is only responsible for not reading response body back, can you provide more details about how it actually breaks the protocol?
And in any case, it's better to make it configurable via the job parameters, not sure which section it would go to thought

Arriven: Actually, if there's a need to not read responses there already a way to do it via packetgen so we can probably just remove this parameter and return to a version where we don't allocate response at all and pass nil to `client.Do` instead. It's not ideal as fasthttp allocates it behind the scenes anyway but at least it'll reduce the amount of code we have to support

src/utils/metrics/prometheus.go
```diff
@@ -233,8 +231,26 @@ func getBasicAuth() (string, string, error) {
 // PushGatewayCA variable to embed self-signed CA for TLS
 var PushGatewayCA string
 
-// getTLSConfig returns tls.Config with system root CAs and embedded CA if not empty
-func getTLSConfig() (*tls.Config, error) {
+// getTLSConfig returns cached tls.Config with system root CAs and embedded CA if not empty
+func getTLSConfig() (config *tls.Config, err error) {
```
Arriven, line 235: If my memory serves me well, this is called only once by `go pushMetrics` or something like that at the start of the program (if it's not than it's a bug)
The key thing is that those metrics were introduced initially to have rough understanding of how many people are using this app to be able to plan attacks better but right now there are so many users that in order to keep up with that I'd need to scale up something like a full Thanos cluster so I'm thinking about removing push metrics completely

Arriven (commented):
I like the general changes but there are some things that need to be changed

src/core/spdnsclient/unixlike_dnsclient.go
```diff
@@ -0,0 +1,570 @@
...
+		}
+		if 'A' <= b && b <= 'Z' {
+			b += 0x20
+		}
+		if a != b {
+			return false
+		}
+	}
+	return true
+}
```
roman-kruglov, line 449: you had something like `equalFoldAscii` in some other file, didn't it have similar logic?

roman-kruglov (commented):
just suggestions, feel free to ignore

src/utils/metrics/prometheus.go
deputinizer: > (if it's not than it's a bug)

Yes, it was called multiple times

src/core/countrychecker/countrychecker.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	return true, country
+}
+
+func fetchLocationInfoFastHTTP(logger *zap.Logger, global *job.GlobalConfig) (country, ip string, err error) {
+	ipCheckerURI := "https://api.myip.com/"
+	requestTimeout := 3 * time.Second
+	maxIdleConns := 10
```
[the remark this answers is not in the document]
deputinizer: Well `http.ClientConfig` required a pointer to a variable, but const can't be made into a pointer.

src/core/countrychecker/countrychecker.go
```diff
@@ -0,0 +1,157 @@
...
+		StaticHost:      nil,
+		TLSClientConfig: nil,
+		Timeout:         &requestTimeout,
+		ReadTimeout:     &requestTimeout,
+		WriteTimeout:    &requestTimeout,
+		IdleTimeout:     &requestTimeout,
+		MaxIdleConns:    &maxIdleConns,
+		ProxyURLs:       "",
+		LocalAddr:       "",
+		Interface:       "",
```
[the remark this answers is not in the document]
deputinizer: Ok

src/core/customresolver/customresolver.go
deputinizer: > as it will work across processes on different machines

Well, custom DNS cilent makes it more universal - for different machines. 
Without this DNS client [from stoppropaganda], golang does not resolve DNS on Android.

src/core/spdnsclient/net.go
deputinizer: It's just copied from golang source, but without reading `/etc/resolv.conf` and Unix things.

src/job/http.go
deputinizer: > If I'm not mistaken this is only responsible for not reading response body back, can you provide more details about how it actually breaks the protocol?

Lol it's obvious that the sever responds with body, so you have to read it...

> it's better to make it configurable via the job parameters

It would be better to just `conn.Close()` instead of having broken connection for the next request with fasthttp.

src/core/countrychecker/countrychecker.go
```diff
@@ -0,0 +1,157 @@
...
+func NewCountryCheckerConfigWithFlags() *CountryCheckerConfig {
+	const maxFetchRetries = 3
+
+	var res CountryCheckerConfig
+
+	flag.StringVar(&res.countryBlackListCSV, "country-list", utils.GetEnvStringDefault("COUNTRY_LIST", "Ukraine"), "comma-separated list of countries")
+	flag.BoolVar(&res.strictCountryCheck, "strict-country-check", utils.GetEnvBoolDefault("STRICT_COUNTRY_CHECK", false),
+		"enable strict country check; will also exit if IP can't be determined")
+	flag.IntVar(&res.maxRetries, "country-check-retries", utils.GetEnvIntDefault("COUNTRY_CHECK_RETRIES", maxFetchRetries),
+		"how much retries should be made when checking the country")
```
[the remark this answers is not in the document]
deputinizer: It's just untouched code from countrychecker inside `utils` folder.

src/core/spdnsclient/unixlike_dnsclient.go
deputinizer: It's not my code. Copied from golang source. Just edited private variables/code to make it use custom provided `resolv.conf`

src/core/spdnsclient/lookup.go
```diff
@@ -0,0 +1,221 @@
...
+	return r.GoLookupHost(ctx, host)
+}
+
+func (r *SPResolver) lookupIP(ctx context.Context, network, host string) (addrs []net.IPAddr, err error) {
+	if true {
+		return r.goLookupIP(ctx, network, host)
+	}
+
+	ips, _, err := r.goLookupIPCNAME(ctx, network, host)
+	return ips, err
```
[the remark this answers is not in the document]
deputinizer: Oh, nice catch. I have to cross-check with Go source.

deputinizer:
Ok since @Arriven commited changes already, I'm closing due to merge conflicts.

src/job/http.go
Arriven: unless http server implements tcp socket itself with the ability to tap into system level acknowledgements there's no way for it to determine whether the client has received the response or not and it can only write responses to it until the network layer buffer is overflown. If you don't care about validity of the code you might generate more load by not waiting for the response, especially if you are doing an L4 attack (the benefit of generating plausible traffic for an L4 attack is questionable but still).

I agree that it's better to have a way to send more correct traffic as it has more chance of reaching the server and I removed this setting (we can still do those L4 attacks with packetgen)

src/core/customresolver/customresolver.go
Arriven: Hmm, the android issue is interesting but if you specify raw IPs in the config (outsourcing the DNS work out of the app) you won't encounter it

Arriven: that being said, I'd probably be ok with having custom dns that would work on android but it would be better to have it as a template function (Ideally having this dns as a separate module with minimum ties to the main package). Template approach has benefits of being more configurable and it would also not require doing any synchronization work as it would be possible to configure it to do only one dns resolution per target in the config

src/core/countrychecker/countrychecker.go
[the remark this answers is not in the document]
Arriven: yup, I saw that a bit later but I just went ahead and used raw fasthttp client instead of using `core/http` in countrychecker. core one has too many features for a simple function like that and also requires passing a lot more parameters

src/utils/metrics/prometheus.go
Arriven: well, not an issue now anyway. the server is dead for couple weeks already so I just removed this code completely :)

src/job/http.go
deputinizer: > you might generate more load by not waiting for the response

Yeah, that's fun too :wink: :stuck_out_tongue_winking_eye:

roman-kruglov:
maybe just rebase, I'm sure there isn't a lot of conflicts

PR #547, opened 2022-05-31
Mobile support (android, ios)

roman-kruglov:
# Description

Made with `fyne` - a terminal emulator which supports text selection and copying; for mobile platforms. Also respects dark and light themes. If you remove the theming code - the terminal is only 65 lines long.

## How to build

```
$ go get fyne.io/fyne/v2/cmd/fyne
$ fyne install
```
or
```
go get github.com/fyne-io/fyne-cross
go get fyne.io/fyne/v2/cmd/fyne
go install fyne.io/fyne/v2/cmd/fyne
```

Then have Java SDK and NDK installed (or just extracted somewhere) and make sure `NDK_HOME` is properly set. To build for all possible android devices you need an NDK with `android16` compilers (at least fyne demanded them from me) - like r19, but I bet there are newer versions which have those too. The newest version of NDK doesn't. But it is also possible to limit cpu architectures and not build all possible variants.

Then just run `fyne package -os android -appID com.some.id`

It is also possible to build a release build with `fyne release` and have it published to PlayStore, but I doubt they will pass it there. Building for IOS is possible too, but I didn't try it yet (described here https://developer.fyne.io/started/mobile).

## Type of change

## TBD

Limit the maximum amount of text logs displayed not to overflow memory.


main.go
```diff
@@ -99,7 +102,14 @@ func main() {
 	metrics.InitOrFail(ctx, logger, *prometheusOn, *prometheusListenAddress, jobsGlobalConfig.ClientID, country)
 
 	reporter := newReporter(*logFormat, logger)
-	job.NewRunner(runnerConfigOptions, jobsGlobalConfig, reporter).Run(ctx, logger)
+	runner := job.NewRunner(runnerConfigOptions, jobsGlobalConfig, reporter)
+
+	if mobile.IsMobile {
```
FPavlovskiy, line 107: I would suggest to use another main_mobile.go to start a runner. In this case it even simpler to do it in separate repo and use 1000n as a dependency, as @arriven proposed before.

roman-kruglov: so you would copy all main's content to it as well?.. imo it's not robust. Or how would you use this main's content? Is it possible to run it as a plain function inside another main?

Initially I tried having an all contained file with `init()` which runs everything without any need to modify main, but unfortunately fyne doesn't allow running its main loop not in the main thread.

FPavlovskiy: What about moving generic part of main to some API-like function, and then call it from other projects like a mobile app?

roman-kruglov: In this case you don't need to support anything - no manifests, no stuff - just a bit of code which introduces something where to show the output. And you just build it for androids (and ios, but I didn't test that).

Anyway, even if a separate repo is to be pursued, we would need to make changes similar to turning the whole app into a "lib" and then having main.go just plug and run it. @arriven are you ok with that? would it be still preferable?

PR #565, opened 2022-08-12
Added flag for periodic GC invocation 

millfreedom:
# Description
Added flag for periodic GC invocation  to reduce overall memory consuption in pooling scenarios (db1000nx100)

## Type of change

## How Has This Been Tested?
manually tested

### DIFF
--- a/main.go
+++ b/main.go
@@ -29,6 +29,7 @@ import (
 	"net/http"
 	pprofhttp "net/http/pprof"
 	"os"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


millfreedom:
@arriven Богдане, пінг.
Гляньте, чи можете прийняти PR.

arriven (approved):
Interesting, I feel like it would be more effective to just tune garbage collection via the GO_GC env variable, but at the same time this flag would allow us to check results more easily

Could also be helpful with issues like https://github.com/arriven/db1000n/issues/555

